PR #4, opened 2025-09-13
docs: Fix a perfectly fine comment :)

ConnorNeed:


src/Cameras/computer_vision/computer_vision/zed_aruco_detector.py
```diff
@@ -62,7 +62,7 @@ def depth_image_callback(self, msg: Image):
         )
 
     def camera_info_callback(self, msg: CameraInfo):
-        # Gets camera intrinsic parameters
```
noodlekid, line 65: Does this method really "set" anything? Consider changing it back to get :)

noodlekid (changes requested):

src/Cameras/computer_vision/computer_vision/zed_aruco_detector.py
ConnorNeed: NO

noodlekid (approved):

PR #8, opened 2025-09-19
refactor: clean up unused code in elevation mapping

ConnorNeed:
Just a straight refactor removing commented out code and un-used code.

Functionallity should not change at all. This is me helping prep for [ROVER-437](https://jira.engsoc.net/browse/ROVER-437) to make it easier to understand

### DIFF
--- a/.devcontainer/jetson/devcontainer.json
+++ b/.devcontainer/jetson/devcontainer.json
@@ -5,9 +5,10 @@
 
   "runArgs": [
     "--network=host",
-    "--volume=/dev:/dev",
+    "--privileged",
     "--volume=/sys:/sys",
     "--volume=/etc/timezone:/etc/timezone",
+    "--volume=/tmp/argus_socket:/tmp/argus_socket",
     "--runtime=nvidia",
     "--gpus=all",
     "--ipc=host",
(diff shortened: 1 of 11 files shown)


ErikCald (approved):
All looks good to me. I will try to test it tomorrow on my Jetson at home.

ConnorNeed:
> All looks good to me. I will try to test it tomorrow on my Jetson at home.

Were you ever able to test it?

ErikCald:
> > All looks good to me. I will try to test it tomorrow on my Jetson at home.
> 
> Were you ever able to test it?

I just tried to run it now. Nothing was being produced the /traversability_map topic but I don't think it's elevation mapping's fault I think it's never launching because of ros2 container shenanigans. 

The shenanigans so far:
 - `ros2 launch navigation navigation.launch.py` fails to launch the zed node, says it launches the zed in the /nav2_container which doesn't make sense since that container is never started since the nav2.launch.py is not set to composable, and also why is zed wanting that container
 - `ros2 launch navigation zed.launch.py` does launch the zed node into the normally named zed container.
 - elevation mapping is supposed to find and attach to zed container so probably related.

I need to look at this more but I ran of time right now.

Side note for later, I needed `sudo apt install ros-humble-grid-map-rviz-plugin`

ConnorNeed:
Tested it and got it running on the jetson. There were some issues with zed and my vs code dev container config. Seems good now though

TomasWilliston (approved):
Looks good to me. Thanks for cleaning it up.

PR #12, opened 2025-10-07
Cicd fixes

ConnorNeed:
Gets the CICD pipeline building for arm64 and amd64 building.

For arm toolchain check out https://github.com/introlab/rtabmap_ros/issues/1288 as they ran into the same issue. I did something very similar.

This pull request introduces several improvements and refactors to the Docker build system for the project, focusing on multi-architecture support, build caching, and workflow flexibility. The main changes include enhanced handling of build arguments and toolchains, improved multi-stage Dockerfile organization, and updates to CI workflow configurations.

**Docker build and multi-architecture improvements:**

* Added support for specifying a custom toolchain file (`aarch64_toolchain.cmake`) during cross-architecture builds, improving compatibility for ARM64 targets. [[1]](diffhunk://#diff-7a8f1fd887a4eb13dfcea508a155c91482b283ea2c4089e03f412a30b16db8ceR1-R18) [[2]](diffhunk://#diff-0b20a7e05df484916805c64c8289e1321376d75b6f2e0a49744e71af8312f4e9L105-R123) [[3]](diffhunk://#diff-0b20a7e05df484916805c64c8289e1321376d75b6f2e0a49744e71af8312f4e9L120-R136)
* Refactored the build scripts (`docker_build.sh`, `gstreamer-config/docker_build.sh`) to support a new `--test` mode, handle persistent buildx builders more flexibly, and allow custom temporary directories for Docker builds. [[1]](diffhunk://#diff-0b20a7e05df484916805c64c8289e1321376d75b6f2e0a49744e71af8312f4e9R15) [[2]](diffhunk://#diff-0b20a7e05df484916805c64c8289e1321376d75b6f2e0a49744e71af8312f4e9R28-R31) [[3]](diffhunk://#diff-0b20a7e05df484916805c64c8289e1321376d75b6f2e0a49744e71af8312f4e9R48-R51) [[4]](diffhunk://#diff-0b20a7e05df484916805c64c8289e1321376d75b6f2e0a49744e71af8312f4e9L75-L83) [[5]](diffhunk://#diff-0b20a7e05df484916805c64c8289e1321376d75b6f2e0a49744e71af8312f4e9R97-R105) [[6]](diffhunk://#diff-6a2186f3280f136936df50ba065398aebbdb63847c8c9a2f9133bf44dfce1693R31-R36) [[7]](diffhunk://#diff-6a2186f3280f136936df50ba065398aebbdb63847c8c9a2f9133bf44dfce1693L68-R72)

[...]

PR #14, opened 2025-10-10
chore: mount keys directly to allow for ssh within remote dev container

ConnorNeed:
Mounting in the SSH agent doesn't seem to work when ssh'ed into another machine. Instead just bind in the ssh keys.


ConnorNeed:
CICD will continue to fail till https://github.com/CPRT/rover/pull/12 is merged in. Not a big deal for this since it changes nothing in terms of docker images

ErikCald (approved):
You have the same change in #8 but good to get this in sooner in case that pr is dragged on a bit

PR #16, opened 2025-10-22
Optimize Decay Computation in radialInflateSerial

jackcwong2007:
- Replaced previous function with a pre-computed lookup table 
- Unsure about what to do with previous function code

### DIFF
--- a/src/Nav/cprt_gridmap_filters/include/PreserveCostInflationFilter.hpp
+++ b/src/Nav/cprt_gridmap_filters/include/PreserveCostInflationFilter.hpp
@@ -67,6 +67,14 @@ public:
                            const Eigen::Vector2d &position, const float value);
 
 private:
+  // function to get decay values
+  float getDecay(uint16_t, uint16_t) const;
+
+  /* Builds the lookup table for decay values if uninitiallized, map resolution
+   * changes, or inflation parameters change
+   */
+  void create_lookup_table(const T &mapOut);
+
   // Input layer name to inflate.
   std::string inputLayer_;
 
@@ -84,6 +92,12 @@ private:
 
   // Decay rate
   double decayRate_;
+
+  // lookup table for decay values
+  std::vector<std::vector<float>> decayTable_;
+
+  // maximum size of decay lookup table (square)
+  uint16_t maxDx_;
 
 }; // class PreserveCostInflationFilter
 
--- a/src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
+++ b/src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
@@ -81,6 +81,28 @@ template <typename T> bool PreserveCostInflationFilter<T>::configure() {
     return false;
   }
 
+  // max size of lookup table
+  maxDx_ = static_cast<int>(std::ceil(decayInflationRadius_));
+  maxDy_ = static_cast<int>(std::ceil(decayInflationRadius_)); 
+
+  // prepare array
+  decayTable_.resize(maxDx_ + 1, std::vector<double>(maxDy_ + 1, 0.0));
+
+  //fill up tables with values
+  for (int dx = 0; dx <= maxDx_; ++dx) {
+    for (int dy = 0; dy <= maxDy_; ++dy) {
+      double distance = std::sqrt(dx * dx + dy * dy);
+      if (distance <= coreInflationRadius_) { // inside inflation radius so no decay
+        decayTable_[dx][dy] = 1.0;
+      } else if (distance <= decayInflationRadius_) { //outside core but inside decay
+        double exponent = -decayRate_ * (distance - coreInflationRadius_);
+        decayTable_[dx][dy] = std::exp(exponent);
+      } else { // too far for any decay
+        decayTable_[dx][dy] = 0.0;
+      }
+    }
+  }
+
   return true;
 }
 
@@ -137,6 +159,25 @@ void PreserveCostInflationFilter<T>::computeWithSimpleSerialMethod(
 }
 
 template <typename T>
+inline double PreserveCostInflationFilter<T>::getDecay(int dx, int dy) const {
+
+  if (dx < 0 || dy < 0) {
+    return 0.0;
+  }  
+  if (dx < dy) {
+    std::swap(dx, dy);
+  }
+  if (dx > maxDx_) {
+    dx = maxDx_;
+  }
+  if (dy > maxDy_) {
+    dy = maxDy_;
+  }
+
+  return decayTable_[dx][dy];
+}
+
+template <typename T>
 void PreserveCostInflationFilter<T>::radialInflateSerial(
     grid_map::GridMap &mapOut, const grid_map::Position &position,
     const float value) {
@@ -155,10 +196,12 @@ void PreserveCostInflationFilter<T>::radialInflateSerial(
     if (distance <= this->coreInflationRadius_) {
       cellValue = std::max(cellValue, value);
     } else if (distance <= this->decayInflationRadius_) {
-      const double decayedValue =
-          value *
-          std::exp(-this->decayRate_ * (distance - this->coreInflationRadius_));
+      int dx = static_cast<int>(
+          std::round(std::abs(cellPosition.x() - position.x()) / mapOut.getResolution()));
+      int dy = static_cast<int>(
+          std::round(std::abs(cellPosition.y() - position.y()) / mapOut.getResolution()));
 
+      const double decayedValue = value * getDecay(dx, dy);
       cellValue = std::max(cellValue, static_cast<float>(decayedValue));
     }
   }
--- a/src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
+++ b/src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
@@ -15,6 +15,7 @@
 
 #include "grid_map_cv/utilities.hpp"
 
+
 namespace grid_map {
 
 template <typename T>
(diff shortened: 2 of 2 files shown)


src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -81,6 +81,28 @@ template <typename T> bool PreserveCostInflationFilter<T>::configure() {
     return false;
   }
 
+  // max size of lookup table
+  maxDx_ = static_cast<int>(std::ceil(decayInflationRadius_));
```
ConnorNeed: Won't maxDx_ and maxDy_ will always be the same? Why do we need 2 varibles?

ConnorNeed: Typically dx and dy will be in units of the number of cells. decayInflationRadius_ is in m. You will need to divide by the resolution to get the correct max number of cells.

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -81,6 +81,28 @@ template <typename T> bool PreserveCostInflationFilter<T>::configure() {
...
+  maxDy_ = static_cast<int>(std::ceil(decayInflationRadius_)); 
+
+  // prepare array
+  decayTable_.resize(maxDx_ + 1, std::vector<double>(maxDy_ + 1, 0.0));
```
ConnorNeed: Where is decayTable defined?

ConnorNeed: I think we can get away with the percision of a float instead of a double. This will cut the memory usage of our lookup table in half.

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -81,6 +81,28 @@ template <typename T> bool PreserveCostInflationFilter<T>::configure() {
...
+
+  //fill up tables with values
+  for (int dx = 0; dx <= maxDx_; ++dx) {
+    for (int dy = 0; dy <= maxDy_; ++dy) {
+      double distance = std::sqrt(dx * dx + dy * dy);
```
ConnorNeed: Watch your units

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -155,10 +196,12 @@ void PreserveCostInflationFilter<T>::radialInflateSerial(
     if (distance <= this->coreInflationRadius_) {
       cellValue = std::max(cellValue, value);
     } else if (distance <= this->decayInflationRadius_) {
-      const double decayedValue =
-          value *
-          std::exp(-this->decayRate_ * (distance - this->coreInflationRadius_));
+      int dx = static_cast<int>(
```
ConnorNeed: This seems to have dx and dy in meters, rounded to the nearest whole meter. See other comments for further explination but this doesn't pass the smell test given how our whole robot is only slightly larger than 1m.

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -136,6 +158,25 @@ void PreserveCostInflationFilter<T>::computeWithSimpleSerialMethod(
                         "NORMAL COMPUTATION TIME = %f", (end - start));
 }
 
+template <typename T>
+inline double PreserveCostInflationFilter<T>::getDecay(int dx, int dy) const {
+
+  if (dx < 0 || dy < 0) {
```
ConnorNeed: If you force the function declaration to take in uint instead of int you can make it the caller's resposibility to ensure what is passed in is non-negative

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -136,6 +158,25 @@ void PreserveCostInflationFilter<T>::computeWithSimpleSerialMethod(
...
+    return 0.0;
+  }  
+  if (dx < dy) {
+    std::swap(dx, dy);
```
ConnorNeed: I really like this choice to swap to ensure that we know which one is bigger all the times. Can you come up with a way we can leverage this to allow for a smaller lookup table?

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -136,6 +158,25 @@ void PreserveCostInflationFilter<T>::computeWithSimpleSerialMethod(
...
+  }
+  if (dx > maxDx_) {
```
ConnorNeed: I like std::max in this scenario. That's solely a readability thing. The compiler will turn it into the same optimized machine code.

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -136,6 +158,25 @@ void PreserveCostInflationFilter<T>::computeWithSimpleSerialMethod(
...
+    dx = maxDx_;
+  }
+  if (dy > maxDy_) {
```
ConnorNeed: same as above

ConnorNeed (changes requested):
Good start. Setting to draft as we work on it together and get it tested.

ConnorNeed:
CI/CD pipeline shows compile errors ... Maybe not everything pushed?

ConnorNeed:
Still seems to be missing the changes to the header file.

jackcwong2007:
> CI/CD pipeline shows compile errors ... Maybe not everything pushed?

Looked through errors and the error message I got is:

ERROR: failed to build: failed to solve: process "/bin/bash -o pipefail -c source /opt/ros/humble/setup.bash     && colcon build --continue-on-error        ${BUILD_FLAGS}" did not complete successfully: exit code: 2

Google didn't return anything and I couldn't find ${BUILD_FLAGS} anywhere. I'll bring my laptop in tomorrow and hopefully we can troubleshoot this.

ConnorNeed:
> > CI/CD pipeline shows compile errors ... Maybe not everything pushed?
> 
> Looked through errors and the error message I got is:
> 
> ERROR: failed to build: failed to solve: process "/bin/bash -o pipefail -c source /opt/ros/humble/setup.bash && colcon build --continue-on-error ${BUILD_FLAGS}" did not complete successfully: exit code: 2
> 
> Google didn't return anything and I couldn't find ${BUILD_FLAGS} anywhere. I'll bring my laptop in tomorrow and hopefully we can troubleshoot this.

Scroll up a bit and it tells you what the compile failures are. The code is referencing data members that don't exist. Look specifically at lines 370-405 in the log messages.

jackcwong2007:
Will take a look, thanks!

src/Nav/cprt_gridmap_filters/include/PreserveCostInflationFilter.hpp
jackcwong2007, line 1: Changes
- Creation and filling of lookup table was moved to bool PreserveCostInflationFilter<T>::update, was not meant to be in configure.
- Moved decayTable_ outside of the function (otherwise is cleared every time the function runs)
- Deleted outdated radialInflateSerial function
- Made changes to header file

Sorry that this is so late!

ConnorNeed:
Failing test was an issue with the build server. It seems to be fixed. I am queuing it up to try run again.

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -14,6 +14,7 @@
 #include <string>
 
 #include "grid_map_cv/utilities.hpp"
+typedef unsigned int uint;
```
ConnorNeed: If you include cstdint you can use uint16_t natively

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -136,6 +165,18 @@ void PreserveCostInflationFilter<T>::computeWithSimpleSerialMethod(
                         "NORMAL COMPUTATION TIME = %f", (end - start));
 }
 
+template <typename T>
+inline float PreserveCostInflationFilter<T>::getDecay(uint dx, uint dy) const {
+  if (dx < dy) {
+    std::swap(dx, dy);
```
ConnorNeed: Why are we swapping?

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -80,10 +81,38 @@ template <typename T> bool PreserveCostInflationFilter<T>::configure() {
...
+  }
+  // prepare table (square)
+  decayTable_.resize(maxDx_ + 1, std::vector<float>(maxDx_ + 1, 0.0));
+
+  // fill up tables with values
+  for (int dx = 0; dx <= maxDx_; ++dx) {
+    for (int dy = 0; dy <= maxDx_; ++dy) {
+      float distance = std::sqrt(dx * dx + dy * dy) * mapOut.getResolution();
+      if (distance <=
+          coreInflationRadius_) { // inside inflation radius so no decay
```
ConnorNeed, line 102:
```suggestion
          coreInflationRadius_) {
```

Comment self explanitory from the code

src/Nav/cprt_gridmap_filters/src/PreserveCostInflationFilter.cpp
```diff
@@ -80,10 +81,38 @@ template <typename T> bool PreserveCostInflationFilter<T>::configure() {
...
+        decayTable_[dx][dy] = 1.0;
+      } else if (distance <=
+                 decayInflationRadius_) { // outside core but inside decay
+        float exponent = -decayRate_ * (distance - coreInflationRadius_);
+        decayTable_[dx][dy] = std::exp(exponent);
+      } else { // too far for any decay
```
ConnorNeed, line 108:
```suggestion
      } else {
```

ConnorNeed (commented):
Almost there!

ConnorNeed:
If you finish cleaning up these last few comments I can do final testing and merge it in tommorrow.

ConnorNeed:
Thanks Jack. I will test tomorrow and get it merged in as long as it works. If you would prefer to be there for the testing then we can do it together either tomorrow or Saturday depending on your schedule.

jackcwong2007:
While I would love to be there for the testing I unfortunately have two midterms on Monday which I will be going into hibernation for. If anything comes up, keep me updated!

ConnorNeed:
Rebased off main

ConnorNeed (approved):

PR #17, opened 2025-10-30
Rover 591

jujuz455:
Refactored the rtcm_node to replace the busy-wait loop with a threading.Thread. 
Fixed high CPU busy-wait

### DIFF
--- a/src/Nav/gps/gps/rtcm_pub_node.py
+++ b/src/Nav/gps/gps/rtcm_pub_node.py
@@ -1,5 +1,6 @@
 import rclpy
 from rclpy.node import Node
+import threading
 from rtcm_msgs.msg import Message as Rtcm
 from pyubx2.ubxtypes_configdb import SET_LAYER_RAM, SET_LAYER_BBR, SET_LAYER_FLASH
 from pyubx2 import (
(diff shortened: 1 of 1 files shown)


.devcontainer/jetson/devcontainer.json
ConnorNeed, line 1: I get why you made these changes to have it be able to run on your mac. However I have a different config file called ./devcontainer/windows/devcontainer.json which should have the same changes you made here. This config is specific for running on the rover computer so I don't want to make these changes to this one.

src/Nav/gps/gps/rtcm_pub_node.py
```diff
@@ -68,11 +85,7 @@ def load_params(self):
         self.persistent = (
             self.get_parameter("Persistent").get_parameter_value().bool_value
         )
-        self.declare_parameter("Freq", 2.0)
-        self.freq = self.get_parameter("Freq").get_parameter_value().double_value
-        if self.freq <= 0:
-            self.get_logger().warn("Frequency must be positive. Defaulting to 2.0 Hz.")
-            self.freq = 2.0
+        # deleted frequency parameter as unused, I/O is handled by an event-driven thread
```
ConnorNeed: Not sure this comment is useful as it only means anything to people who knew the old code

src/third-party/kindr
ConnorNeed, line 1: Uncommit this change as I am guessing it was unintensional

build.sh
```diff
@@ -10,4 +10,4 @@ find ./src -path ./src/third-party -prune -o \
 
 rosdep install --from-paths src -i -r -y
 
-colcon build --symlink-install --continue-on-error --cmake-args=-DCMAKE_BUILD_TYPE=Release --parallel-workers $(nproc)
\ No newline at end of file
+colcon build --symlink-install --continue-on-error --cmake-args=-DCMAKE_BUILD_TYPE=Release --parallel-workers 1
```
ConnorNeed: I get computers with less ram and only do one package at a time but I would like to leave the default as $(nproc) which returns the number of CPU cores present. Maybe a good compermise would be add a optional agrument which allows the called to set the number of workers?

src/Nav/gps/gps/rtcm_pub_node.py
```diff
@@ -121,23 +134,60 @@ def config_rtcm(self, port_type: str = "USB"):
...
+                # the threading will sleep here until data is available (0% CPU usage)
+                # or the port is closed by destroy_node
+
+                raw, parsed_data = self.serial_conn.read()
+
+                if not raw:
+                    # for read timeouts or empty reads, just continue
+
+                    if rclpy.ok():
+                        self.get_logger().warn("Serial read returned no data.")
```
ConnorNeed: Maybe downgrade this to an info statement as it isn't necessarity a bad thing.

src/Nav/gps/gps/rtcm_pub_node.py
```diff
@@ -121,23 +134,60 @@ def config_rtcm(self, port_type: str = "USB"):
...
+                msg = Rtcm()
+                msg.message = list(raw)
+
+                self.rtcm_pub.publish(msg)
+                self.get_logger().debug(
+                    f"Published RTCM message of length {len(raw)}. Parsed message: {parsed_data}"
+                )
+        except Exception as e:
+
+            # Catch-all for any exceptions in the read loop
```
ConnorNeed: Code here is self explanitory. Doesn't need a comment

src/Nav/gps/gps/rtcm_pub_node.py
ConnorNeed:
```suggestion
                        self.get_logger().info("Serial read returned no data.")
```

src/Nav/gps/gps/rtcm_pub_node.py
ConnorNeed: Remove this no longer existing parameter from https://github.com/CPRT/rover/blob/main/src/Nav/gps/launch/heading.launch.py#L26

ConnorNeed (commented):
Really good work!

ConnorNeed (approved):
LGTM :)

PR #18, opened 2025-11-01
Refactor Servo Architecture Part 1

Jayden442:
This is the first part of the task, which is the actual refactoring of the servo package. The second part is to update Teleop control to be compatible with this.

- Redesigned the servo architecture to use inheritance with a common class and config yaml file.
- Config files are made to be as similar as possible
- Now uses publisher/subscriber as opposed to services
- The client is the publisher, the server is the subscriber
- Degrees are now radians.

Tested by having the client send 2 unique positions. 

Tested by running usb servo and client from the same launch file but it is commented out right now just as it was before.

Part 2 (The joystick part) has been merged.

Task: [ROVER-411](https://jira.engsoc.net/browse/ROVER-411)


ConnorNeed:
Thanks Jayden. I will take a look through tomorrow. We will hold off on merging it in until part 1 and 2 are both ready to be merged.

src/HW-Devices/servo_pkg/config/pi_controller.yaml
```diff
@@ -1,16 +1,17 @@
 pi_Servo_node:
   ros__parameters:
-    servos_used: 2
+    # this is the highest servo number
+    max_num_servo: 1
```
ConnorNeed: Seems unnecessarily confusing rather than just using the number of servos?

src/HW-Devices/servo_pkg/launch/usb_servo_launch.launch.py
```diff
@@ -6,21 +6,25 @@
...
         [
             launch_ros.actions.Node(
                 package="servo_pkg",
                 executable="USB_Servo",
                 name="USB_Servo_node",
-                parameters=[parameters_file],
+                parameters=[parent_params, child_params],
             ),
-            # for example client
+            # if you wanted to run client as well
```
ConnorNeed: We can probably remove this. People can launch it using ros2 run cmds if they want to test.

ConnorNeed (commented):
Looks good. Just need to wait for the porting of all the control to use the new system before merging

ConnorNeed:
Looks like you have to rebase off main and resolve the conflicts. Once you do there will be a pylint config .pylintrc. 
If you run:
```bash
pylint src
```
It will run some basic checks to ensure the changes are compliant with our codebase pylint rules. They are not super strict (yet) but they help catch some stuff.

ConnorNeed (approved):

PR #21, opened 2025-11-07
Nav launch

ConnorNeed:
## Config changes:
1) Changing pub_resolution to OPTIMIZED (the default) increases the throughput a massive amount and makes the whole system much more stable. Not sure if you have any experience with this setting but it might be worth giving it a try.

## Launch changes
1) Removed bringup.launch.py since we have never used this file
2) Renamed container_name launch argument in the nav2 launch file as it was conflicting with the container_name launch argument in the zed file. Since they are both getting declared when running navigation.launch.py they must be unique
3) Fixed typo from transversability to traversability
4) Set defaults to use only zed since we are in URC mode

## Other changes
1) Added ros.sh which you can source using ". ros.sh" which is shorter to type
2) Updated readme 
3) Mount zed folder so we don't have to constantly re-download and optimize the models


ConnorNeed:
@jackcwong2007 These were the changes that had nothing to do with your code but caused us not to be able to test your stuff on Wednsday. Once this gets merged in you will have to rebase your branch off of main again.

.devcontainer/jetson/devcontainer.json
```diff
@@ -11,6 +11,7 @@
     "--volume=/sys:/sys",
     "--volume=/etc/timezone:/etc/timezone",
     "--volume=/tmp/argus_socket:/tmp/argus_socket",
+    "--volume=/usr/local/zed:/usr/local/zed",
```
ErikCald, line 14: This will likely cause the docker container to not launch if the /usr/local/zed directory doesn't exist on the host system. Since it's only the jetson dev container, that will only be anytime we wipe and reflash the jetson. Maybe we need to add a line to the first time install to create the directory?

src/Nav/navigation/config/zed/zed_override_params.yaml
```diff
@@ -6,7 +6,7 @@
             grab_resolution: 'HD720' # The native camera grab resolution. 'HD2K', 'HD1080', 'HD720', 'VGA', 'AUTO'
             grab_frame_rate: 60 # ZED SDK internal grabbing rate
 
-            pub_resolution: "CUSTOM" # The resolution used for output. 'NATIVE' to use the same `general.grab_resolution` - `CUSTOM` to apply the `general.pub_downscale_factor` downscale factory to reduce bandwidth in transmission
+            pub_resolution: "OPTIMIZED" # The resolution used for output. 'NATIVE' to use the same `general.grab_resolution` - `CUSTOM` to apply the `general.pub_downscale_factor` downscale factory to reduce bandwidth in transmission
```
ErikCald, line 9: I didn't know about this so I did some research. OPTMIZED shows up in the changelogs as a new option 9 months ago. But I looked through the code of the zed ros2 wrapper and the code that loads this parameter is only checking for NATIVE and CUSTOM, it doesn't even check for OPTIMIZED. 

I'm curious if you will see this is the logs:
` * Publishing resolution: NATIVE -> Fix the value in YAML!`

Or something like `Not valid 'general.pub_resolution' value: '%s'. Using default NATIVE setting instead.`

ErikCald (changes requested):

src/Nav/navigation/config/zed/zed_override_params.yaml
ConnorNeed: https://github.com/CPRT/rover/blob/main/src/third-party/zed-ros2-wrapper/zed_wrapper/config/common_stereo.yaml#L24

It's their default setting in their config file

.devcontainer/jetson/devcontainer.json
ConnorNeed: You have to install the zed sdk on the host in order to get the udev rules to work properly anyways. I forgot to add that to the setup script. See new commit.

Source: https://gist.github.com/adujardin/2d5ce8f000fc6a7bd40bee2709749ff8

src/Nav/navigation/config/zed/zed_override_params.yaml
ConnorNeed: https://github.com/CPRT/rover/blob/main/src/third-party/zed-ros2-wrapper/zed_components/src/zed_camera/src/zed_camera_component.cpp#L870

.devcontainer/jetson/devcontainer.json
ConnorNeed: Thanks for catching this

src/Nav/navigation/config/zed/zed_override_params.yaml
ErikCald: Oh yeah it does have optimized in what we have cloned. I was looking at a main branch on the zed wrapper repo, they must have weird branch versioning.  Ok well hopefully it makes things better

ErikCald (approved):
LGTM 
image

ConnorNeed:
Thanks Erik. If the optimized setting gives issues with later processing we can always change it back but with it turned on I was able to get 25fps so I really like that part.

ErikCald:
> Thanks Erik. If the optimized setting gives issues with later processing we can always change it back but with it turned on I was able to get 25fps so I really like that part.

Cool. 25fps is fast! Now we just need elevation mapping to operate at 25fps hehe

ErikCald:
They made a few different changes to this pub resolution setting over the past year.
I found this commit from feb 14 where they removed OPTIMIZED:
https://github.com/stereolabs/zed-ros2-wrapper/commit/0f9557ff8edf7fca8db5bc08c0705ae07a3e9b9a#diff-6b1e40243c6ad2b72c35d3eaf286ecbf7cd56574b9fffc2a253285c740a600fa

At some point we might want to try out ZED SDK 5.0, but also what we have seems to work fine and maybe using what we've tested already is better. They have a tag on the ros2 wrapper for [humble-v5.0.0](https://github.com/stereolabs/zed-ros2-wrapper/releases/tag/humble-v5.0.0)

ConnorNeed:
I hate that they don't give any reasoning behind the change. https://github.com/stereolabs/zed-ros2-wrapper/pull/304/files

OPTIMIZED seems to have the same result as setting their new parameter to COMPACT. I am good with upgrading the wrapper and switching to using the COMPACT parameter but that can be a seperate PR.

In terms of moving to the 5.0 SDK I will leave that to your discretion on whether that is something you want to take on. Theoretically it shouldn't be hard to migrate but I expect there will be serious roadblocks.

PR #22, opened 2025-11-07
fix: get working

ConnorNeed:
1) Re-adds the radialInflateSerial function which is necessary.
2) Only updates the lookup table if necessary
3) Makes getDecay private (Principle of least privilage)


ConnorNeed:
@jackcwong2007 Feel free to ask any questions. Once you have had a chance to look it over, go ahead and merge it into your branch

PR #23, opened 2025-11-11
Navigation tuning

ConnorNeed:
## Changed how we launch autonomy:
slam.launch.py:
- zed camera
- all localization
- elevation/traversability mapping
- Ouster (default no)

nav2.launch.py
- all nav2 components

When launching nav2 as a composible nodes there is a race condition with the zed camera. Always launch slam.launch.py before nav2.launch.py

## Removed layers from output of traversability map
- Makes the output smaller so the dds doesn't work as hard

## Tested with infill filter
- Seems to work
- Lower latency than the box blur method
- Need to verify that the output is at least as good

### DIFF
--- a/src/Nav/localization/launch/ekf.launch.py
+++ b/src/Nav/localization/launch/ekf.launch.py
@@ -36,7 +36,7 @@ def launch_setup(context):
     # This function is executed at launch time
 
     use_sim_time = LaunchConfiguration("use_sim_time").perform(context)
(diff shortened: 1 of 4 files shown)


ErikCald (approved):
Yeah looks good to me.  So now nav2 is launched on it's own, and then everything is launched by slam.launch.py.

ConnorNeed:
> Yeah looks good to me. So now nav2 is launched on it's own, and then everything is launched by slam.launch.py.

Correct

PR #24, opened 2025-11-13
Aruco changes

ConnorNeed:
For https://jira.engsoc.net/browse/ROVER-672

1. Move the aruco tag detector to the new video_streaming package
2. Add mounting to dev container config to match the jetson and allow to work on cameras from any linux host
3. Use RGB buffers (what we use in our pipeline because it is nvidia's preference) to reduce the number of converversions. Aruco converts to grayscale (from BGR) inside the aruco detection anyways so we can do it outside without issue.
4. Detect and draw using gstreamer inplace transformation reducing copies further.


.devcontainer/amd64/devcontainer.json
```diff
@@ -5,7 +5,9 @@
 
   "runArgs": [
     "--network=host",
-    "--volume=/dev:/dev",
+    "--privileged",
```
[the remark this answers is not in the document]
ConnorNeed: No

.devcontainer/amd64/devcontainer.json
```diff
@@ -5,7 +5,9 @@
...
+    "--volume=/dev/v4l/by-id:/dev/v4l/by-id",
+    "--volume=/dev/serial/by-id:/dev/serial/by-id",
```
ErikCald, line 10: My understanding is that --privileged gives the container access to the host filesystem so that it can lookup the symlinks in /dev/v4l/by-id?

ErikCald:
So you feed RGB images to the aruco and it converts it to greyscale assuming its a BGR, but since its greyscale and we are detecting a black/white aruco it doesn't matter. Seems fine to be assuming my understanding is correct.

ErikCald (approved):
I haven't worked with aruco in cpp so I don't have much comments on the cpp itself, cpp looks all syntactically good to me!
Gemini_Generated_Image_70rx1a70rx1a70rx

.devcontainer/amd64/devcontainer.json
ErikCald: Nevermind I'm good I found this info on --privileged:
- Device Visibility: It forces Docker to populate the container's /dev/ directory with all the devices found on the host. Without this (or a specific --device flag), the container's /dev/video0 would not exist.
- Permission: It bypasses the cgroup rules that usually prevent containers from reading/writing to hardware.

ConnorNeed:
> So you feed RGB images to the aruco and it converts it to greyscale assuming its a BGR, but since its greyscale and we are detecting a black/white aruco it doesn't matter. Seems fine to be assuming my understanding is correct.

Not quite. RGB to greyscale would be a different conversion than BGR to greyscale (the RGB are not equal weights). If I were to pass the RGB plane into the aruco detector the first thing it would do it convert to greyscale assumeing it is BGR. That's why I convert to greyscale properly on this line (https://github.com/CPRT/rover/pull/24/files#diff-e4431fec4adc15e8316e5850c7c396980e7ee140c3bf4d82ad06ff00607712d8R53) and pass in the greyscale.

ConnorNeed:
The one thing that does treat the RGB as BGR is the drawing of the box. So technically the bounding box will be a different colour. It just so happens that the bounding box is green and therefore it looks the same. Even if this wasn't true I still wouldn't care since all it would do is change the colour of the BB.

jujuz455 (approved):
LGTM!

ConnorNeed:
@jujuz455 This is a little on the advanced side of gstreamer since it implements our own element but take a look through and feel free to ask any questions.

PR #25, opened 2025-11-14
[DRAFT] Add Lidar 2D SLAM EKF for indoor autonomy testing

ErikCald:
Created an efk profile and a lidar_2dslam.launch.py which will launch all the regular autonomy stuff as normal, except it will launch the ouster lidar and use a different efk profile that just replaces the gps odom with a pose from slam_toolbox. That way the system operates very closely to how it will with proper gps, and we can performance test the elevation mapping pipeline without worrying about the ouster potentially interfering with it (which other solutions might have that problem).

JIRA: https://jira.engsoc.net/browse/ROVER-673

# Copliot summary
This pull request introduces a new EKF (Extended Kalman Filter) profile for lidar-based 2D SLAM localization and adds configuration support for it throughout the launch and configuration files. Additionally, it provides a tailored parameter file for `slam_toolbox` to optimize SLAM performance with lidar data. The main changes are grouped below:

**EKF profile support for lidar-based 2D SLAM:**

* Added new configuration files `lidar_2dslam_ekf_local.yaml` and `lidar_2dslam_ekf_global.yaml` for local and global EKF settings specific to lidar-based 2D SLAM. These files include sensor topics, frame settings, and covariance parameters optimized for lidar and IMU integration. [[1]](diffhunk://#diff-9480c44d67d9250c0b6b0a57e5cd46d9be686f942eed5f8876f63d1b1e78f02bR1-R143) [[2]](diffhunk://#diff-291ef85090b9fd4d8bc0d67af7bccc3a16e39339309020ae2fb22363f9e0ef5bR1-R143)
* Updated `ekf.launch.py` to support selection of the new `lidar_2dslam` profile, including changes to profile constants, default handling, and error messages. [[1]](diffhunk://#diff-54e86a4399a7251e8bf3a09e3788a12a283d354086e29008fbe2d1404575a0edL26-R65) [[2]](diffhunk://#diff-54e86a4399a7251e8bf3a09e3788a12a283d354086e29008fbe2d1404575a0edL110-R119)

**Launch file integration for EKF profile selection:**

* Modified `localization.launch.py` to propagate the `ekf_profile` argument, allowing users to select between `gps` and `lidar_2dslam` EKF profiles at launch time. This includes argument declaration, passing, and default handling. [[1]](diffhunk://#diff-977c51d8627a0283fc211ebf7e865c4f2120b863b769c2b1ed8a34e278677019R23) [[2]](diffhunk://#diff-977c51d8627a0283fc211ebf7e865c4f2120b863b769c2b1ed8a34e278677019R40-R44) [[3]](diffhunk://#diff-977c51d8627a0283fc211ebf7e865c4f2120b863b769c2b1ed8a34e278677019L77-R86) [[4]](diffhunk://#diff-977c51d8627a0283fc211ebf7e865c4f2120b863b769c2b1ed8a34e278677019R116)

**SLAM toolbox configuration for lidar mapping:**

* Added a new parameter file `slam_toolbox_params.yaml` with settings tuned for lidar-based online mapping, including solver configuration, scan matching, keyframe creation, loop closure, and penalty parameters. These changes are designed to improve mapping accuracy and robustness for lidar-based SLAM.

### DIFF
--- a/src/Nav/navigation/launch/lidar_2dslam.launch.py
+++ b/src/Nav/navigation/launch/lidar_2dslam.launch.py
@@ -51,11 +51,11 @@ def generate_launch_description():
                     {
                         "target_frame": "lidar_link",
                         "transform_tolerance": 0.01,
-                        "min_height": -0.3,  # ouster is 0.425m above ground
+                        "min_height": -0.1,  # ouster is 0.425m above ground
                         "max_height": 0.1,
                         "angle_min": -3.14159,
                         "angle_max": 3.14159,
-                        "angle_increment": 0.01227184630308513,  # ouster in 512x10 mode so use 2*pi/512
+                        "angle_increment": 0.006135923,  # ouster in 1024x10 mode so use 2*pi/1024
                         "scan_time": 0.1,  # matches 10 Hz
                         "range_min": 1.5,  # avoids accidentally picking up the rover itself
                         "range_max": 20.0,
@@ -79,7 +79,7 @@ def generate_launch_description():
             slam_params_file,
             {
                 "use_sim_time": use_sim_time,
-                "publish_tf": False,  # False: Disable slam_toolbox publishing odom and map tf frames
+                "publish_tf": True,  # False: Disable slam_toolbox publishing odom and map tf frames
             },
         ],
         remappings=[
@@ -129,8 +129,8 @@ def generate_launch_description():
             declare_slam_params_file,
             slam_container,
             ouster_cmd,
-            slam_toolbox_node,
-            # slam_cmd,
-            # localization_cmd,
+            # slam_toolbox_node,
+            slam_cmd,
+            localization_cmd,
         ]
     )
--- a/src/Nav/navigation/launch/lidar_2dslam.launch.py
+++ b/src/Nav/navigation/launch/lidar_2dslam.launch.py
@@ -119,6 +119,7 @@ def generate_launch_description():
         launch_arguments={
             "use_sim_time": use_sim_time,
             "ekf_profile": "lidar_2dslam",
+            "launch_gps": "false",
         }.items(),
     )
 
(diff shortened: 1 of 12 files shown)


src/Nav/navigation/launch/lidar_2dslam.launch.py
```diff
@@ -0,0 +1,135 @@
...
+    slam_container = ComposableNodeContainer(
+        name="slam_container",
+        namespace="",
+        package="rclcpp_components",
+        executable="component_container",
+        composable_node_descriptions=[
+            # Pointcloud to LaserScan conversion
+            ComposableNode(
+                package="pointcloud_to_laserscan",
+                plugin="pointcloud_to_laserscan::PointCloudToLaserScanNode",
```
ErikCald, line 44: I think I maybe did something weird by converting the 3d pointcloud from the ouster into a laser scan instead of just giving the laser scan from the ouster to the slam toolbox. This does let me configure a bit more about what ranges points to accept and min/max threshold, but now that I think about it there must exist a laser scan trimming package that is better than doing this.

ConnorNeed: Slam toolbox has min_laser_range and max_laser_range.

ConnorNeed:
A few general concerns:
1. Is slam toolbox so good that it can completely replace the ZED odom suite. Maybe so but I think that's probably asking too much. Can we incorporate something the the ~/odom topic from zed to give us some more data? This should also fix the need for initial odom publisher.
2. I am not a huge fan of the inital odometry publisher. It feels like a hackier hack than necessary. Maybe it's the only way but I think we can find a better way. 
3. Nothing is being used for z position that I can see. This has hurt us in the past.

ErikCald:
Never ended up being worth using the lidar for indoor testing. And we now can just test outside in more realistic conditions so this is no longer necessary.

PR #26, opened 2025-11-15
Ros2imagesrc

ConnorNeed:
For https://github.com/CPRT/rover/pull/26

This pull request adds a new GStreamer plugin for ROS2 image streaming to the project. The main changes introduce a new `ros2image` plugin, including its build configuration, header, and registration code, enabling GStreamer pipelines to source images from ROS2 topics.

**Addition of ROS2 image streaming plugin:**

* Added `ros2image` subdirectory to the GStreamer plugins build (`gst-plugins/CMakeLists.txt`), enabling compilation of the new plugin.
* Created `gst-plugins/ros2image/CMakeLists.txt` to configure build dependencies for GStreamer, ROS2, and sensor messages, and to install the plugin correctly.

**Implementation of plugin source element:**

* Introduced `gstros2src.hpp`, defining the `GstRos2ImageSrc` class and its internal queueing and ROS2 subscription logic for images and compressed images.
* Added `plugin.cpp` to register the new `ros2src` element with GStreamer as a plugin, specifying metadata and initialization logic.

### DIFF
--- a/src/Cameras/video_streaming/gst-plugins/CMakeLists.txt
+++ b/src/Cameras/video_streaming/gst-plugins/CMakeLists.txt
@@ -1,3 +1,4 @@
 # gst-plugins/CMakeLists.txt
 
 add_subdirectory(aruco)
+add_subdirectory(ros2image)
--- a/src/Cameras/video_streaming/gst-plugins/ros2image/CMakeLists.txt
+++ b/src/Cameras/video_streaming/gst-plugins/ros2image/CMakeLists.txt
@@ -1,0 +1,50 @@
+# gst-plugins/ros2image/CMakeLists.txt
+
+find_package(PkgConfig REQUIRED)
+find_package(rclcpp REQUIRED)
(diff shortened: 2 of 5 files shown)


ErikCald (approved):
I don't think we have anyone on the team who can actually review and give good comments to all this gstreamer cpp apart from you @ConnorNeed  lol. So... LGTM!

There are some function declarations in the cpp file that feel like they belong in a hpp file, but this looks like gstreamer plugin stuff which might have different coding standards so I defer to you.

ConnorNeed:
> There are some function declarations in the cpp file that feel like they belong in a hpp file, but this looks like gstreamer plugin stuff which might have different coding standards so I defer to you.

Yeah it's a mess in that regard. Gstreamer is a c library so you typically use a c style with minimal header files. But obviously I need c++ for the ros2 libraries so you end up with this mix of both.

jujuz455 (approved):
LGTM! (sorry it took me long time to google stuff

PR #29, opened 2025-11-19
updating limits with encoder movement

Darren-Wallace:
Rechecked and updated 2 of the limits. Should be good to go now until it moves again 🫠


ConnorNeed (approved):
Thanks Darren

PR #30, opened 2025-11-20
Rover 394

DeclanMcIntaggart:
Remade the config files for the gps to factory reset before and changed the frequency of gps/fix and the heading


src/Nav/gps/gps/rtcm_pub_node.py
ConnorNeed, line 1: iirc, there is some configuration that happens in this file. Given that we now configure externally we should remove that logic,

src/Nav/gps/gps/rtcm_pub_node.py
```diff
@@ -85,12 +68,20 @@ def load_params(self):
...
+        if self.freq <= 0:
+            self.get_logger().warn("Frequency must be positive. Defaulting to 2.0 Hz.")
+            self.freq = 2.0
 
         self.declare_parameter("Baudrate", 38400)
         self.baudrate = (
             self.get_parameter("Baudrate").get_parameter_value().integer_value
         )
-        self.declare_parameter("Device", "/dev/ttyACM0")
+        self.declare_parameter(
```
ConnorNeed: This is just the default path. It can be overwriten in the launch file. Using /dev/ttyACM0 makes sense as a default.

src/Nav/gps/launch/heading.launch.py
```diff
@@ -5,16 +5,6 @@
 
 
 def generate_launch_description():
-    config_dir = os.path.join(get_package_share_directory("gps"), "config")
```
ConnorNeed, line 8: Maybe add a launch config with a default of false to load config on startup.

src/Nav/gps/launch/rover.launch.py
```diff
@@ -14,14 +14,6 @@ def generate_launch_description():
 
     params_file = os.path.join(config_dir, "gps.yaml")
 
-    Rover_config_file = os.path.join(config_dir, "Rover_config.ubx")
```
ConnorNeed, line 15: Same here

src/Nav/gps/config/fr_basestation.py
```diff
@@ -0,0 +1,176 @@
...
+    serial_out.write(ubx.serialize())
+
+
+def config_rtcm(port_type: str) -> UBXMessage:
+    """
+    Configure which RTCM3 messages to output.
+    """
+
+    print("\nFormatting RTCM MSGOUT CFG-VALSET message...")
+    layers = 7  # 1 = RAM, 2 = BBR, 4 = Flash (can be OR'd)
```
ConnorNeed: Create some named constants to remove this magic number

src/Nav/gps/config/fr_basestation.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def config_svin(port_type: str, acc_limit: int, svin_min_dur: int) -> UBXMessage:
+    """
+    Configure Survey-In mode with specied accuracy limit.
+    """
+
+    print("\nFormatting SVIN TMODE CFG-VALSET message...")
+    tmode = TMODE_SVIN
+    layers = 1
```
ConnorNeed: Why only layer 1?

src/Nav/gps/config/fr_basestation.py
```diff
@@ -0,0 +1,176 @@
...
+    transaction = 0
+    acc_limit = int(round(acc_limit / 0.1, 0))
```
ConnorNeed: Why are we multiplying accuracy limit by 10? Also that is an odd way to multiple an integer by 10.

src/Nav/gps/config/fr_basestation.py
```diff
@@ -0,0 +1,176 @@
...
+            # configure either Survey-In or Fixed Timing Mode
+            if TMODE == TMODE_SVIN:
+                msg = config_svin(PORT_TYPE, ACC_LIMIT, SVIN_MIN_DUR)
+
+            send_msg(stream, msg)
+    with Serial(
+        "/dev/ttyACM0",
+        38400,
+    ) as stream:
+        layers = 7  # RAM + BBR + Flash for persistent storage
```
ConnorNeed: Same concern as above

src/Nav/gps/config/fr_basestation.py
```diff
@@ -0,0 +1,176 @@
...
+            msg = config_rtcm(PORT_TYPE)
+            send_msg(stream, msg)
+
+            # configure either Survey-In or Fixed Timing Mode
+            if TMODE == TMODE_SVIN:
+                msg = config_svin(PORT_TYPE, ACC_LIMIT, SVIN_MIN_DUR)
+
+            send_msg(stream, msg)
+    with Serial(
+        "/dev/ttyACM0",
```
ConnorNeed: This will be a problem

src/Nav/gps/config/fr_basestation.py
```diff
@@ -0,0 +1,176 @@
...
+        38400,
+    ) as stream:
+        layers = 7  # RAM + BBR + Flash for persistent storage
+        transaction = 0
+        cfg_data = [
+            ("CFG_UART1_BAUDRATE", 9600),
```
ConnorNeed: Why this baudrate?

src/Nav/gps/config/fr_heading.py
```diff
@@ -0,0 +1,205 @@
...
+    Send config message to receiver.
+    """
+    print("Sending configuration message to receiver...")
+    print(ubx)
+    serial_out.write(ubx.serialize())
+
+
+def config_usb_relposned() -> UBXMessage:
+    """
+    Configure NAV-RELPOSNED to output on USB at 1Hz.
```
ConnorNeed: I thought we were doing 5hz?

src/Nav/gps/config/fr_heading.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    print("\nFormatting USB NAV-RELPOSNED MSGOUT CFG-VALSET message...")
+    layers = 7  # RAM + BBR + Flash for persistent storage
+    transaction = 0
+    cfg_data = [
+        ("CFG_MSGOUT_UBX_NAV_RELPOSNED_USB", 1),  # 1 = 1 time per epoch
```
ConnorNeed: What's the conversion from epoch to hz?

ConnorNeed (changes requested):
Here are some things to get started with.

I will challenge you to see if you can create a helper python library that contains all the shared helper functions reducing code dupliations.

Given that the configuration scripts are going to be used as command line tools, use argparse (https://docs.python.org/3/library/argparse.html) to allow us to change some of the parameters from the command line. Make sure to use whatever the "Correct" versions for our setup as the defaults.

src/Nav/gps/config/fr_heading.py
DeclanMcIntaggart: It is 1/rate_ms, where the rate should be in seconds and NAV_RATE_MS, so a value of NAV_RATE_MS is 200 is equal to 5 hz

src/Nav/gps/config/fr_heading.py
DeclanMcIntaggart: I forgot to change the comments

src/Nav/gps/config/fr_basestation.py
DeclanMcIntaggart: That was just the number that I am pretty sure that saw when I was looking at the base station before doing anything to it

ConnorNeed:
Before you start worring about the new task of reading the base station via a different interface this needs to get cleaned up and merged in.

src/Nav/gps/config/fr_basestation.py
DeclanMcIntaggart: The original one we have aka acc_limit is in terms of mm but CFG_TMODE_SVIN_ACC_LIMIT needs it in terms of 0.1mm. I was just mimicing what pyubx2 had in their code for this bit so I just copied their. I think they were doing the divide by 0.1 to try to show that it is meant to be in the units of 0.1 mm rather then 1mm

src/Nav/gps/launch/heading.launch.py
```diff
@@ -28,8 +29,20 @@ def generate_launch_description():
                     {
                         "Device": "/dev/serial/by-id/usb-FTDI_FT230X_Basic_UART_D30I1LY5-if00-port0"
                     },
+                    # expose launch arg to node; node can act on it to load config
```
ConnorNeed: This was obviously AI but please double check the work. This doesn't make any sense in the context of the PR. The part right below it runs the config. Why are we passing it into the node?

src/Nav/gps/launch/rover.launch.py
```diff
@@ -32,14 +28,35 @@ def generate_launch_description():
         executable="ublox_gps_node",
         output="both",
         remappings=ublox_remappings,
-        parameters=[params_file],
+        parameters=[params_file, {"load_config": load_config}],
```
ConnorNeed: Same issue

src/Nav/gps/launch/rover.launch.py
```diff
@@ -32,14 +28,35 @@ def generate_launch_description():
...
             os.path.join(
                 get_package_share_directory("gps"), "launch", "heading.launch.py"
             )
-        )
+        ),
+        launch_arguments={"load_config": load_config}.items(),
     )
 
-    return launch.LaunchDescription([ublox_gps_node, heading_cmd])
+    return launch.LaunchDescription(
```
ConnorNeed, line 28: Why is some of this declared above and some of this in the return statement? Either style is fine but please pick one

ConnorNeed (commented):
Much better. Good work

ConnorNeed (approved):
LGTM

PR #36, opened 2025-11-24
Feature/srt component

jujuz455:
This PR implements the SrtNode for video streaming using GStreamer. 

**Key Changes**

1. Registered SrtNode as a component using RCLCPP_COMPONENTS_REGISTER_NODE.
Updated CMakeLists.txt to build a shared library instead of an executable.

2. Removed Custom Interfaces: Deleted Srtstat.msg and SetStreamingParams.srv to avoid creating new interface definitions.
ROS 2 Parameters: Implemented on_parameter_change callback.

3. Cleaned up package.xml and CMakeLists.txt (removed rosidl generation and unused dependencies).
    Updated video_streaming.launch.py to load video_streaming::SrtNode into the composable node container.
    Fixed build dependencies for input_node (restored interfaces package link).

Notes
Implemented a new test_mode parameter to support local development.
Updated the default SRT port from 12345 to 7001 (standard convention) in both the C++ node and launch file.

### DIFF
--- a/src/Cameras/video_streaming/CMakeLists.txt
+++ b/src/Cameras/video_streaming/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.16)
-project(video_streaming)
+project(video_streaming LANGUAGES CXX)
 
 # ---- Build settings
 set(CMAKE_CXX_STANDARD 17)
@@ -8,24 +8,20 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 # ---- Dependencies
 find_package(ament_cmake REQUIRED)
-find_package(rosidl_default_generators REQUIRED)
-find_package(rosidl_default_runtime REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_components REQUIRED)
-find_package(std_msgs REQUIRED)
+find_package(std_srvs REQUIRED)
 find_package(interfaces REQUIRED)
-# find_package(std_srvs REQUIRED)
 find_package(PkgConfig REQUIRED)
 
-# 1. GStreamer and GLib
 pkg_check_modules(GSTREAMER REQUIRED IMPORTED_TARGET gstreamer-1.0)
 pkg_check_modules(GST_APP REQUIRED IMPORTED_TARGET gstreamer-app-1.0)
 pkg_check_modules(GST_VIDEO REQUIRED IMPORTED_TARGET gstreamer-video-1.0)
-pkg_check_modules(GLIB REQUIRED IMPORTED_TARGET glib-2.0)
 
+add_subdirectory(gst-plugins)
 
-# ---- Build Libraries
-add_library(video_streaming_nodes SHARED
+# ---- Targets
+add_library(${PROJECT_NAME} SHARED
   src/base_video_node.cpp
   src/input_node.cpp
   src/detect_node.cpp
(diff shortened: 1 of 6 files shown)


src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+namespace video_streaming {
+
+SrtNode::SrtNode(const rclcpp::NodeOptions &options)
+    : BaseVideoNode("srt_node", options) {
+  RCLCPP_INFO(this->get_logger(), "Initializing SrtNode...");
+
+  // Start up
+  // latency , iframe_interval
+  this->declare_parameter<std::string>("srt_uri", "srt://127.0.0.1:12345");
```
ConnorNeed: Typically port 7001 is used as the default for srt.

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -0,0 +1,211 @@
...
+  gchar *desc =
+      g_strdup_printf("interpipesrc listen-to=detect ! "
+                      "nvvidconv ! "
+                      "nvv4l2av1enc name=av1_enc ! "
+                      "rtpav1pay ! queue ! "
+                      "srtsink name=srt_sink uri=%s latency=%d mode=1",
+                      srt_uri.c_str(), latency_val);
+
+  /* // [MAC TEST MODE]
+    gchar *desc = g_strdup_printf(
```
ConnorNeed: Add a boolean parameter called test_mode that can allow you to use this without it being commented out code.

src/Cameras/video_streaming/CMakeLists.txt
```diff
@@ -8,63 +8,78 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 # ---- Dependencies
 find_package(ament_cmake REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
+find_package(rosidl_default_runtime REQUIRED)
```
ConnorNeed: Are these still required?

src/Cameras/video_streaming/package.xml
```diff
@@ -3,16 +3,33 @@
...
+  <build_depend>gstreamer-1.0</build_depend>
+  <build_depend>glib-2.0</build_depend>
+
+  <exec_depend>rclcpp</exec_depend>
+  <exec_depend>rclcpp_components</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
+  <exec_depend>std_srvs</exec_depend>
+  <!-- <exec_depend>rosidl_default_runtime</exec_depend> -->
+  <exec_depend>gstreamer-1.0</exec_depend>
+  <exec_depend>glib-2.0</exec_depend>
```
ConnorNeed: Things that are both build and runtime dependacies can just be added as "depend"

src/Cameras/video_streaming/package.xml
```diff
@@ -3,16 +3,33 @@
...
+
+  <!-- <member_of_group>rosidl_interface_packages</member_of_group> -->
```
ConnorNeed: I don't think we need this.

src/Cameras/video_streaming/include/video_streaming/srt_node.hpp
```diff
@@ -0,0 +1,62 @@
...
+   * Kept for reference in case specific service control is needed later.
+   * =========================================================
+   */
+  // void on_set_params(
+  //   const std::shared_ptr<video_streaming::srv::SetStreamingParams::Request>
+  //   request,
+  //   std::shared_ptr<video_streaming::srv::SetStreamingParams::Response>
+  //   response);
+  // rclcpp::Service<video_streaming::srv::SetStreamingParams>::SharedPtr
+  // params_srv_;
```
ConnorNeed: We can probably remove this. If we do end up going back to it, it won't take very long to add back at all.

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -0,0 +1,211 @@
...
+ * * Per feedback, we are currently pausing this functionality to focus on the
+ * video
+ * * pipeline and using ROS 2 Parameters for control.
+ * * Once custom .msg and .srv files are added to the interfaces package,
+ * * uncomment this block and the corresponding headers.
+ * ==================================================================================
+ */
+
+/*
+// void SrtNode::on_srt_stats(GstElement * element, GstStructure * stats,
```
ConnorNeed: I am obviously not a fan of commented out code lol. Can we just store this function elsewhere until we get it working and merge it in in it's own PR?

src/Cameras/video_streaming/CMakeLists.txt
ConnorNeed, line 1: I think we can undo most of these changes since we are not adding the msg/srv generators.

ConnorNeed (commented):
Great work

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+bool SrtNode::create_pipeline() {
+  std::string srt_uri = this->get_parameter("srt_uri").as_string();
+  int latency_val = this->get_parameter("latency").as_int();
+
+  bool test_mode = false;
+  try {
+    test_mode = this->get_parameter("test_mode").as_bool();
+  } catch (...) {
+    test_mode = this->declare_parameter<bool>("test_mode", false);
```
ConnorNeed: Can we declare the parameter in the constructor to avoid this try catch?

src/Cameras/video_streaming/launch/video_streaming.launch.py
```diff
@@ -34,6 +34,16 @@ def generate_launch_description():
         parameters=[],
     )
 
+    srt_node = ComposableNode(
+        package="video_streaming",
+        plugin="video_streaming::SrtNode",
+        name="srt_node",
+        namespace="",
+        parameters=[
+            {"srt_uri": "srt://127.0.0.1:7001", "latency": 100, "iframe_interval": 30}
```
ConnorNeed: I would say it make more sense for the rover (aka "Sender") to be an SRT listener. That way we can choose to have multiple callers and it doesn't matter what IP the address of the client is. I believe all you have to do is switch the uri to not include the ip address and just have the port.
```suggestion
            {"srt_uri": "srt://:7001", "latency": 100, "iframe_interval": 30}
```

ConnorNeed (commented):
We can test on the rover tonight

ConnorNeed (approved):
Well done

PR #38, opened 2025-12-01
Srt_Stats.msg

jujuz455:
SRT streaming statistics and environment adaptivity to the video_streaming package. 

Update:

**Implemented Stats Publishing, Frequency Control**

I've addressed all the review comments and here is a summary of the changes:

Features & Fixes:

SRT Statistics: Implemented robust parsing for srtsink stats. It now correctly handles the nested structure (Listener mode vs. Caller mode) using standard GST_VALUE_HOLDS_ARRAY checks to prevent crashes.

Frequency Control: Added a stats_frequency parameter to control the publishing rate.

Standard Key-Unit Event: Switched  from g_signal_emit to the standard gst_video_event_new_downstream_force_key_unit API 

For testing the encoder on the rover : I've verified the changes locally and everything runs correctly in the dev container. since I'm away from campus, I'd appreciate it if you could give it a try whenever it's convenient!


src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -28,6 +31,13 @@ SrtNode::SrtNode(const rclcpp::NodeOptions &options)
       "~/trigger_iframe", 10,
       std::bind(&SrtNode::on_iframe_trigger, this, std::placeholders::_1));
 
+  srt_stats_pub_ =
+      this->create_publisher<interfaces::msg::SrtStats>("~/srt_stats", 10);
+
+  srt_stats_timer_ =
+      this->create_wall_timer(std::chrono::milliseconds(1000),
```
ConnorNeed: Can we add a parameter for the frequency this publishes at? 1hz is a good default.

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -135,8 +145,62 @@ void SrtNode::on_bitrate_received(const std_msgs::msg::Int32::SharedPtr msg) {
 void SrtNode::on_iframe_trigger(const std_msgs::msg::Empty::SharedPtr msg) {
   (void)msg;
   if (av1_encoder_) {
-    g_signal_emit_by_name(av1_encoder_, "force-IDR", NULL);
-    RCLCPP_INFO(this->get_logger(), "I-Frame triggered");
+    GstEvent *event = gst_video_event_new_downstream_force_key_unit(
```
ConnorNeed, line 156: The nvv4l2enc plugins used to only respond to a custom type of signal to trigger an keyframe. It looks like last feb a user added a patch that should make this work. We will have to test to make sure our version of the encoder is new enough that it has this patch.

https://forums.developer.nvidia.com/t/nvv4l2h264enc-nack-pli-fir-event-handling/319543/19

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -135,8 +145,62 @@ void SrtNode::on_bitrate_received(const std_msgs::msg::Int32::SharedPtr msg) {
...
+
+    // 3. Packets Sent
+    int64_t val_64 = 0;
+    if (gst_structure_get_int64(stats, "packets-sent", &val_64)) {
+      msg.packets_sent = val_64;
+    }
+
+    // 4. Packets Lost
+    int val_int = 0;
+    if (gst_structure_get_int(stats, "packets-lost", &val_int) ||
```
ConnorNeed: Seems like it is called `packets-sent-lost`?

Here is the source code:
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-bad/ext/srt/gstsrtobject.c#L2036

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -135,8 +145,62 @@ void SrtNode::on_bitrate_received(const std_msgs::msg::Int32::SharedPtr msg) {
...
+        gst_structure_get_int(stats, "pkt-snd-loss-total", &val_int)) {
+      msg.packets_lost = val_int;
+    }
+
+    // 5. Packets Retransmitted
+    int ret_int = 0;
+    if (gst_structure_get_int(stats, "packets-retransmitted", &ret_int) ||
+        gst_structure_get_int(stats, "pkt-ret", &ret_int)) {
```
ConnorNeed: Just 'packets-retransmitted' should work. Don't think we need 'pkt-ret'

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -135,8 +145,62 @@ void SrtNode::on_bitrate_received(const std_msgs::msg::Int32::SharedPtr msg) {
...
+  if (!srt_sink_)
+    return;
+
+  GstStructure *stats = nullptr;
+  g_object_get(srt_sink_, "stats", &stats, NULL);
+
+  if (stats) {
+    auto msg = interfaces::msg::SrtStats();
+    msg.header.stamp = this->now();
+
```
ConnorNeed, line 179: When srtsink is in listener mode (Which we will be most of the time on the rover) the stats struct actually contains an array of GstStructures for each of the callers. Add an if statement here to see is there is a element in the structure called "callers". If there is then we want to grab the first element in that array (most of the time there will only be one) and then that should be the same format as the GstStructure that you are reading here. Make sure to handle your de-referencing correctly to avoid memory leaks.

Some links I think you might find useful:
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-bad/ext/srt/gstsrtobject.c#L2086
https://docs.gtk.org/gobject/struct.ValueArray.html
https://docs.gtk.org/gobject/method.Value.take_boxed.html
https://gstreamer.freedesktop.org/documentation/gstreamer/gststructure.html?gi-language=c

ConnorNeed (commented):
Good work. Let me know if you have any questions!

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -135,11 +153,84 @@ void SrtNode::on_bitrate_received(const std_msgs::msg::Int32::SharedPtr msg) {
...
+                (GstStructure *)gst_value_get_structure(first_caller);
+          }
+        }
+      }
+    }
+
+    // 1. RTT: seconds
+    int rtt_ms = 0;
+    if (gst_structure_get_int(target_stats, "rtt-ms", &rtt_ms)) {
+      msg.rtt = static_cast<double>(rtt_ms) / 1000.0;
```
ConnorNeed: We can probably leave this value in ms. Expected rtt is between 10-100ms and is easer to view than 0.01 and 0.1

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -135,11 +153,84 @@ void SrtNode::on_bitrate_received(const std_msgs::msg::Int32::SharedPtr msg) {
...
+    auto msg = interfaces::msg::SrtStats();
+    msg.header.stamp = this->now();
+
+    GstStructure *target_stats = stats;
+
+    // Listener mode: check if "callers" field exists
+    if (gst_structure_has_field(stats, "callers")) {
+      const GValue *callers_val = gst_structure_get_value(stats, "callers");
+
+      if (callers_val && GST_VALUE_HOLDS_ARRAY(callers_val)) {
```
ConnorNeed: Disclaimer: This is a really nitpicky comment and is more from a teaching perspective. I typically am not this picky for this codebase otherwise there are some members that would never be able to get anything approved.

Guard clauses (aka the early return pattern) can reduce the number of indentations and make the logic cleaner. 

It works by checking the error condition first. If the error is wrong, free resources and return. Then for the rest of the function you don't have to worry about that error case.

Any time you are nested inside more than 3 indents it is typically a "Bad Smell" that there might be a cleaner way to do it.

You have a good guard clause on line 168-169. You can also use one on lines 174, 184, 186, and 191.

ConnorNeed (commented):
Nice job! My last exam is on Wednsday so I can help you test it on the rover Thursday, Friday or Saturday. 

Another side note: Typically you don't reuse git branches after a merge. It just makes the git history and rebasing more of a pain. Going forward just create a new branch off of main for every time you merge in then start working on something new.

src/Cameras/video_streaming/src/srt_node.cpp
jujuz455: Thanks for the detailed explanation! I've updated the code to use early returns.

ConnorNeed (approved):
Approved. Hold off on merging until after it is tested on the rover. Make sure to use "Squash and Merge" to keep the git history cleaner and in this case it will also save you merge conflicts.

PR #40, opened 2025-12-11
Getting ready for Odrives interface

ConnorNeed:
For https://jira.engsoc.net/browse/ROVER-678

Creates a base motor controller class to allow for us to use multiple (Odrive + talon) motor controllers in the same system.

Also fixes a bug where inverted encoder joints causes the arm to have the wrong start up value.


Darren-Wallace (approved):
Looks good to me. I'm good with sticking with the BaseWrapper name.

PR #41, opened 2025-12-11
feat: deepstream deps and functional bottle model

ConnorNeed:
No rush to review it. I know it is exams :)

Now runs with gst-launch. I will integrate into our detect application in a different pull request.

Screenshot from 2025-12-11 10-12-57


src/Cameras/video_streaming/config/bottle/bottle.txt
```diff
@@ -0,0 +1,38 @@
...
+
+# Model files
+model-engine-file=bottle.jetson.engine
+onnx-file=bottle_detector.onnx
+
+network-type=0                  # detector
+infer-dims=3;640;640
+batch-size=1
+process-mode=1
+network-mode=0
```
jujuz455: I noticed that the tickets for rock pick and orange hammer require FP16 (network-mode=2) to meet the ≥5 FPS target on the Orin, should this also be FP16?

ConnorNeed: Really great catch. I was able to get it to run at 30fps even with 32bit precision but you are right the orin is optimized for 16bit precision so when we get our real custom model we should try both and compare the resource usage and precision of both modes.

ConnorNeed:
I set Network Mode to fp16 for faster perfomance, run on every frame instead of every other frame and only show one detection per frame.

ErikCald (approved):

ConnorNeed:
Rebased off main. I will merge once build passes

PR #45, opened 2025-12-18
feat: gradient filter

ConnorNeed:
This filter offers 2 estimation methods that can be used for determinding the slope: Sobel and Central Difference.

There are also 2 different output modes: Squared and Magnitude. The only difference is after calculating the squared gradient the mag version takes the square root to give it in absolute units. Squared mode is slightly mode efficient with one less sqrt operation but mag mode is closer to what we currently have. 

The values are similar but slightly different to the current slope calculations but much faster. We will have to re-tune the weightings for the new rover. 

Sobel with Magnitude output is very similar to our current output. Central Difference may need a secondary smoothing filter. 

Goes from 17hz -> 18hz and I expect it to scale better than current in larger maps better than our current setup.

### DIFF
--- a/src/Nav/cprt_gridmap_filters/src/GradientFilter.cpp
+++ b/src/Nav/cprt_gridmap_filters/src/GradientFilter.cpp
@@ -1,0 +1,194 @@
+#include "GradientFilter.hpp"
+
+#include <cmath>
+#include <limits>
+#include <pluginlib/class_list_macros.hpp>
+
+namespace grid_map {
+
+static inline bool isFinite(float value) { return std::isfinite(value); }
+static inline float quietNaN() {
+  return std::numeric_limits<float>::quiet_NaN();
+}
+
+bool GradientFilter::configure() {
+  grid_map::ParameterReader param_reader(this->param_prefix_,
+                                         this->params_interface_);
+  if (!readParam<std::string>(param_reader, "input_layer", input_layer_)) {
+    return false;
+  }
+  if (!readParam<std::string>(param_reader, "output_layer", output_layer_)) {
+    return false;
+  }
+  std::string stencil_str;
+  if (!readParam<std::string>(param_reader, "stencil", stencil_str)) {
+    return false;
+  }
+  if (stencil_str == "central" || stencil_str == "cen") {
+    stencil_ = Stencil::CentralDifference;
+  } else if (stencil_str == "sob" || stencil_str == "sobel") {
+    stencil_ = Stencil::Sobel;
+  } else {
+    RCLCPP_ERROR(this->logging_interface_->get_logger(),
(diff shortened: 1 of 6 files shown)


ConnorNeed:
https://jira.engsoc.net/browse/ROVER-593

src/Nav/cprt_gridmap_filters/src/GradientFilter.cpp
```diff
@@ -0,0 +1,194 @@
...
+        const float h_ml = height_layer(row, col - 1);
+        const float h_mr = height_layer(row, col + 1);
+        const float h_ll = height_layer(row + 1, col - 1);
+        const float h_lc = height_layer(row + 1, col);
+        const float h_lr = height_layer(row + 1, col + 1);
+
+        const bool all_neighbors_finite =
+            (isFinite(h_ul) && isFinite(h_uc) && isFinite(h_ur) &&
+             isFinite(h_ml) && isFinite(h_mr) && isFinite(h_ll) &&
+             isFinite(h_lc) && isFinite(h_lr));
```
ErikCald, line 159: This requires all neighbors to be finite which might be a bit limiting for areas of the pointcloud that are beyond 5 meters where the zed only has a line of points (alot like a lidar pointcloud). The zed pointcloud is very dense for anything closer than 5 meters so its not something we need to consider for awhile

src/Nav/cprt_gridmap_filters/src/GradientFilter.cpp
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  const auto &height_layer = map_in[input_layer_];
+  auto &gradient_layer = map_out[output_layer_];
+
+  const int num_rows = height_layer.rows();
+  const int num_cols = height_layer.cols();
+  const float cell_size_m = static_cast<float>(map_in.getResolution());
+
+  for (int row = 0; row < num_rows; ++row) {
```
ErikCald, line 79: This looks like it could work just fine multi-threaded to if we need to make it scale even better. 
`#pragma omp parallel for collapse(2)` should just work. I think right now our map is too small for this to do anything though.

ErikCald (approved):
Looks good to me
image

src/Nav/cprt_gridmap_filters/src/GradientFilter.cpp
ConnorNeed: This runs after some form of infill filter (either the smoothing filter or opencv infill filter which is now working) so it should be fine even if there are a few holes in the zed data.

src/Nav/cprt_gridmap_filters/src/GradientFilter.cpp
ConnorNeed: I'll let you experiment with this as you are tuning it but a cool thought

PR #47, opened 2026-01-04
New urdf

ConnorNeed:
- Gets the combined (arm + chassis) URDF running in Rviz simulation
- Port over servo node config to use position control
- Uses primiative shape collision checking for faster computation (Includes a python app to help)
- Decimated the STL files in blender for faster RVIZ load (Visually looks almost the same but around 1/10th of the verticies)

Note:
- Some weird things going on with the Chassis URDF but I will fix those when I integrate the swerve drive system


MaybeWilli (approved):
Long as it launches fine, looks good to me

ConnorNeed:
Rebased off main. Going to hold off on merging this in until I get the go ahead to completely decommision Spike's arm. This is a breaking change.

PR #48, opened 2026-01-06
 PLI Back-off & Framerate Control 

jujuz455:
Stacked on top of PR46.

**Change(line259):** 
Now sends an upstream event to the srt_sink.

**New features:**

**Packet Loss Indication (PLI) with Back-off:**

- Implemented an exponential back-off algorithm (doubling delay) for requesting I-Frames when packet loss is detected.

- Prevents flooding the encoder with Key Unit requests during unstable network conditions.

- Configurable parameters: backoff_max_delay_ms and backoff_reset_ms.

**Dynamic Framerate Control:**

- Added support for runtime parameter updates for target_framerate.

- Videorate now updates dynamically without restarting the node when the parameter changes.

### DIFF
--- a/src/Cameras/video_streaming/src/srt_node.cpp
+++ b/src/Cameras/video_streaming/src/srt_node.cpp
@@ -13,12 +13,25 @@ SrtNode::SrtNode(const rclcpp::NodeOptions &options)
     : BaseVideoNode("srt_node", options) {
   RCLCPP_INFO(this->get_logger(), "Initializing SrtNode...");
 
+  // Set last trigger time to 1 hour ago so the first error triggers immediately
+  backoff_state_.last_trigger_time =
+      std::chrono::steady_clock::now() - std::chrono::hours(1);
+  backoff_state_.last_loss_time = std::chrono::steady_clock::now();
+  backoff_state_.last_total_dropped_pkts = 0;
+  backoff_state_.current_delay_ms = 200;
+
   this->declare_parameter<std::string>("srt_uri", "srt://:7001");
   this->declare_parameter<int>("latency", 100);
   this->declare_parameter<int>("iframe_interval", 0);
   this->declare_parameter<bool>("test_mode", false);
   this->declare_parameter<double>("stats_frequency", 1.0);
   this->declare_parameter<int>("target_framerate", 30);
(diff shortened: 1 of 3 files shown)


src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -222,16 +255,69 @@ void SrtNode::on_bitrate_received(const std_msgs::msg::Int32::SharedPtr msg) {
...
+  auto now = std::chrono::steady_clock::now();
+
+  if (current_total_dropped <= backoff_state_.last_total_dropped_pkts) {
+    auto time_since_loss =
+        std::chrono::duration_cast<std::chrono::milliseconds>(
+            now - backoff_state_.last_loss_time)
+            .count();
+
+    if (time_since_loss > backoff_state_.reset_ms) {
+      if (backoff_state_.current_delay_ms != 200) {
```
ConnorNeed: Can we combine these if statements?

ConnorNeed: 200 is a bit of a magic number here. I am assuming this is the starting delay. Can we create a constexpr in the header file to hold this value?

src/Cameras/video_streaming/src/srt_node.cpp
```diff
@@ -222,16 +255,69 @@ void SrtNode::on_bitrate_received(const std_msgs::msg::Int32::SharedPtr msg) {
...
+
+  backoff_state_.last_total_dropped_pkts = current_total_dropped;
+  backoff_state_.last_loss_time = now;
+
+  auto time_since_trigger =
+      std::chrono::duration_cast<std::chrono::milliseconds>(
+          now - backoff_state_.last_trigger_time)
+          .count();
+
+  if (time_since_trigger < backoff_state_.current_delay_ms) {
```
ConnorNeed: Add a RCLCPP_DEBUG here. Might help with debugging in the future but also serves as self documenting code as to why we are returning

ConnorNeed (approved):
Fantastic work! Approving so you can merge it into your other branch if you like but take a look at the comments. They are really small stuff that should only take a couple minutes to clean up. We will test tomorrow!

PR #50, opened 2026-01-11
Fix urdf

ConnorNeed:
Fixes the URDF to have the links in the proper locations. Re-uses STL files between some similar parts.


Darren-Wallace (approved):
Took a quick look and it seems good to me.

PR #51, opened 2026-01-13
SRT client application

jujuz455:
Stack on top of PR 46

### DIFF
--- a/src/Cameras/video_streaming/client_app/srt_client.c
+++ b/src/Cameras/video_streaming/client_app/srt_client.c
@@ -1,0 +1,160 @@
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <gst/gst.h>
+
+
+typedef struct
+{
+  GMainLoop *loop;
+  GstElement *pipeline;
+  guint bus_watch;
+  guint io_watch_id;
+} GlobalData;
+
+static gboolean handle_bus_msg (GstBus * bus, GstMessage * msg, GlobalData * data);
+static gboolean io_callback (GIOChannel * io, GIOCondition condition, GlobalData * data);
+
+int
+main (int argc, char *argv[])
+{
+  GlobalData data;
+  GIOChannel *io = NULL;
+  GstBus *bus;
+  GError *error = NULL;
+
+  /* Initialize GStreamer */
+  gst_init (&argc, &argv);
+
+  if (argc < 2) {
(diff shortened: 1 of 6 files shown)


src/Cameras/video_streaming/client_app/srt_client.c
```diff
@@ -0,0 +1,160 @@
...
+
+
+// srtsrc -> av1parse -> decodebin -> videoconvert -> sink
+  const char *default_uri = "srt://:7001?mode=listener";
+  const char *uri = (argc >= 2) ? argv[1] : default_uri;
+
+  g_print ("Trying to listen on URI: %s\n", uri);
+
+  gchar *pipeline_str = g_strdup_printf (
+      "srtsrc uri=\"%s\" latency=200 ! "
```
ConnorNeed, line 44: As weird as this sounds, I would set the latency to 0ms. Not because this is a reasonable number but the actual (called negotiated) latency is the higher of the sender and reciever latency. Since the sender latency is being controlled by the operator, this should be 0 so it is never taken into consideration.

ConnorNeed (commented):
We will have to add a build system. We have a couple options:
1) Convert to c++ and wrap it in a ros2 node so we can use the ros2 launch system (ease of use and clean but means we need ros2 dependancies)
2) Add it to the cmake file and make use of the ros2 sourcing via a hook (bit hacky but would work and it can stay as a c file)
3) Create a new repo and build it into it's own application (Clean, but will require a completely new repo. Bonus is in the future we could add Windows and/or Mac support and allow now software members to build and run it directly off their machines as well)


I personally am good with 1 or 3. We can talk about it tonight :)

jujuz455:
> We will have to add a build system. We have a couple options:
> 
> 1. Convert to c++ and wrap it in a ros2 node so we can use the ros2 launch system (ease of use and clean but means we need ros2 dependancies)
> 2. Add it to the cmake file and make use of the ros2 sourcing via a hook (bit hacky but would work and it can stay as a c file)
> 3. Create a new repo and build it into it's own application (Clean, but will require a completely new repo. Bonus is in the future we could add Windows and/or Mac support and allow now software members to build and run it directly off their machines as well)
> 
> I personally am good with 1 or 3. We can talk about it tonight :)

jujuz455:
> 1. Convert to c++ and wrap it in a ros2 node so we can use the ros2 launch system (ease of use and clean but means we need ros2 dependancies)

Could you elaborate more on this option? I'm not entirely clear on how to set that up

PR #52, opened 2026-01-17
Swerve controller

ConnorNeed:
Adds a swerve controller and ros2 control system for chassis.

Left to do for chassis to work:
1. Import and configure odrive ros2 package
2. Seperate the options for launching urdf (arm, chassis, both) - currently always both
3. Continue to clean up launch files

Documentation AI generated.


src/URDF/rover_urdf/config/ros2_controllers.yaml
```diff
@@ -0,0 +1,52 @@
...
+      rear_left:
+        steering_angle_name: Left_back_wheel_arm_joint
+        wheel_velocity_name: Left_back_wheel_joint
+      rear_right:
+        steering_angle_name: Right_back_wheel_arm_joint
+        wheel_velocity_name: Right_back_wheel_joint
+    wheel_base: 0.5
+    track_width: 0.4
+    wheel_radius: 0.1
+    odom_publish_rate: 10.0
```
ErikCald: I think 10hz might be too slow, the ekf could use a faster rate. The defaults are 50 hz in the code so im wondering if this is also supposed to be 50 hz here.

ErikCald (approved):
I looked over things and didn't find anything wrong. I'll take another peak at it tomorrow

src/URDF/rover_urdf/config/ros2_controllers.yaml
ConnorNeed: Thanks Erik. I will make this change and merge it in in the morning

PR #53, opened 2026-01-17
feat: SRT client 

jujuz455:
**Refactored the SRT client app. 
It is now a C++ Node with ROS 2 dependencies configured in CMake.
Please let me know if this aligns with the requirements.**

Summery:

Ported the client logic to C++.

Wrapped it in a ROS 2 node

Created a launch file (srt_client.launch.py)

Updated CMakeLists.txt

### DIFF
--- a/.devcontainer/base-station/devcontainer.json
+++ b/.devcontainer/base-station/devcontainer.json
@@ -1,0 +1,48 @@
+{
+  "name": "CPRT Dev Container",
+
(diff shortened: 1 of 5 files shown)


ConnorNeed:
Yep that meets the requirements. We will have to test the pipeline before we merge.

ConnorNeed (approved):
Looks good to me

PR #55, opened 2026-01-19
Box blur dev

omeletteee:
What:
Created a new box blur filter that can be applied to a gridmap layer to assist the rover in determining the traversability of terrain by smoothing over noise and holes. 

Why:
The built-in box blur function the rover had been using previously was inefficient, so this new algorithm was created with the intention of reducing latency. 

How:
This increased efficiency is achieved by using a separable algorithm in place of the naive implementation that was used previously. The gridmap layer inputLayer_ of mapIn is first blurred along the horizontal axis, and then blurred again, vertically before being returned in mapOut as outputLayer_. The function also takes in a blur radius (in units of cells, not distance), which is typically set to one but can be expanded.

Testing:
This code has been tested using the Jetson and the Zed camera, and the results have been compared against the old algorithm to find that they produce highly similar results (with the only differences being found along the edge of the map depending on what mode the old algorithm's edge handling was set to).

Other Notes:
Any potentially unclear lines _should_ be commented in the code, but do let me know if there's anything I need to clarify.
In the future, it would be possible to use weighted averages during the blurring process in order to achieve a Gaussian blur as opposed to a box blur (unweighted) were that to become necessary.

### DIFF
--- a/src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
+++ b/src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
@@ -51,6 +51,8 @@ public:
   
   void VerticalBoxBlur(const Eigen::MatrixXf &layerIn,
                          Eigen::MatrixXf &layerOut, double radius);
+  
+  void MatrixMultiply(const T &mapH, const T &mapV, T &mapOut);
 
 private:
   //! Radius to take the mean from.
--- a/src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
+++ b/src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
@@ -8,9 +8,9 @@
 
 #ifndef CPRTGRIDMAPFILTERS_BOXBLURFILTER_HPP_
 #define CPRTGRIDMAPFILTERS_BOXBLURFILTER_HPP_
-#include <string>
-
+#include <Eigen/Dense>
 #include <filters/filter_base.hpp>
+#include <string>
 
 #include <string>
 #include <vector>
@@ -45,6 +45,12 @@ public:
    * layer.
    */
   bool update(const T &mapIn, T &mapOut) override;
+
+  void HorizontalBoxBlur(const Eigen::MatrixXf &layerIn,
+                         Eigen::MatrixXf &layerOut, double radius);
+  
+  void VerticalBoxBlur(const Eigen::MatrixXf &layerIn,
+                         Eigen::MatrixXf &layerOut, double radius);
 
 private:
   //! Radius to take the mean from.
--- a/src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
+++ b/src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
@@ -1,0 +1,62 @@
+/**
+ * PreserveCostInflationFilter.hpp
+ *
+ *    Created on: Oct 4th, 2025
+ *        Author: Lauren Spargo
+ *  Organization: Carleton Planetary Robotics Team
+ */
+
+#ifndef CPRTGRIDMAPFILTERS_BOXBLURFILTER_HPP_
+#define CPRTGRIDMAPFILTERS_BOXBLURFILTER_HPP_
+#include <string>
+
+#include <filters/filter_base.hpp>
+
+#include <string>
+#include <vector>
+
+namespace grid_map {
+
+/*!
+ * Filter class to find the mean of the values inside a radius.
+ */
+template <typename T> class BoxBlurFilter : public filters::FilterBase<T> {
+public:
+  /*!
+   * Constructor
+   */
+  BoxBlurFilter();
+
+  /*!
+   * Destructor.
+   */
+  virtual ~BoxBlurFilter();
+
+  /*!
+   * Configures the filter from parameters on the Parameter Server
+   */
+  bool configure() override;
+
+  /*!
+   * Computes for each value in the input layer the mean of all values in a
+   * radius around it Saves this mean in an additional output layer.
+   * @param mapIn grid map containing the input layer.
+   * @param mapOut grid map containing the layers of the input map and the new
+   * layer.
+   */
+  bool update(const T &mapIn, T &mapOut) override;
+
+private:
+  //! Radius to take the mean from.
+  double radius_;
+
+  //! Input layer name.
+  std::string inputLayer_;
+
+  //! Output layer name.
+  std::string outputLayer_;
+};
+
+} // namespace grid_map
+
+#endif // CPRTGRIDMAPFILTERS_PRESERVECOSTINFLATIONFILTER_HPP_
--- a/src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
+++ b/src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
@@ -47,18 +47,19 @@ public:
   bool update(const T &mapIn, T &mapOut) override;
 
   void HorizontalBoxBlur(const Eigen::MatrixXf &layerIn,
-                         Eigen::MatrixXf &layerOut, double radius);
-  
-  void VerticalBoxBlur(const Eigen::MatrixXf &layerIn,
-                         Eigen::MatrixXf &layerOut, double radius);
-  
-  //void MatrixMultiply(const T &mapH, const T &mapV, T &mapOut);
+                         Eigen::MatrixXf &layerOut, int r);
 
-  //void MatrixMultiply(const Eigen::MatrixXf &layerHorz, const Eigen::MatrixXf &layerVert, Eigen::MatrixXf &layerOut) 
+  void VerticalBoxBlur(const Eigen::MatrixXf &layerIn,
+                       Eigen::MatrixXf &layerOut, int r);
+
+  // void MatrixMultiply(const T &mapH, const T &mapV, T &mapOut);
+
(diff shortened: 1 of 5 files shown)


src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+template <typename T> bool BoxBlurFilter<T>::configure() {
+  ParameterReader param_reader(this->param_prefix_, this->params_interface_);
+
+  if (!param_reader.get(std::string("radius"), radius_)) {
+    RCLCPP_ERROR(this->logging_interface_->get_logger(),
+                 "BoxBlur filter did not find parameter `radius`.");
+    return false;
+  }
+
+  if (radius_ < 0) {
```
ConnorNeed:
```suggestion
  if (radius_ <= 0) {
```

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+                 "Layer %s does not exist. Unable to apply box blur"
+                 "filter.",
+                 this->inputLayer_.c_str());
+    return false;
+  }
+
+  mapOut.add(this->outputLayer_);
+  mapOutHorz.add(this->outputLayer_);
+
+  HorizontalBoxBlur(mapIn[inputLayer_], mapOutHorz[outputLayer_], radius_); //Blurs input layer from mapIn horizontally, saves blurred values to mapOutHorz on outputLayer_
```
ConnorNeed: Unnecessary comment. If future coders need to understand what that function does they should look at the header file

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+  VerticalBoxBlur(mapOutHorz[outputLayer_], mapOut[outputLayer_], radius_); //Blurs outputLayer_ from mapOutHorz vertically, saves blurred values to mapOut on outputLayer_
```
ConnorNeed: Same here

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+  //Box blur complete!
```
ConnorNeed: It is generally assumed that when a function returns it is complete.

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+{
+  int height = layerOut.rows();
+  int width = layerOut.cols();
+
+  for (int y = 0; y < height; y++) {
+    for (int x = 0; x < width; x++) {
+      //ensuring that max and min cells in radius do not exceed dimensions of gridmap layer
+      int minCoeff = std::max(0, x - r);
+      int maxCoeff = std::min(width - 1, x + r);
+      int numHoles = 0;
```
ConnorNeed: Generally you want to count what you are interested in. We don't really care about the number of holes. We care about the number of valid values. I would rename this to be numValues.

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+      double sum = 0;
+
+      for (int i = minCoeff; i <= maxCoeff; i++) {
+        const float value = layerIn(i, y);
+        if (!std::isfinite(value)) //checking for hole
+        {
+          numHoles += 1; //if hole found, disregard it and continue
+          continue;
+        }
+        sum += value;
```
ConnorNeed:
```suggestion
        if (std::isfinite(value))
        {
          sum += value;
          numValues++;
        }
```

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+      }
+      //calculate and write average to layerOut
+      layerOut(x, y) = sum / (maxCoeff - minCoeff + 1 - numHoles);
```
ConnorNeed:
```suggestion
      if (numValues > 0) {
        layerOut(x, y) = sum / numValues ;
      }
```

We want to avoid divide by zero errors.

src/Nav/cprt_gridmap_filters/include/BoxBlurFilter.hpp
```diff
@@ -0,0 +1,73 @@
...
+   * @param mapIn grid map containing the input layer.
+   * @param mapOut grid map containing the layers of the input map and the new
+   * layer.
+   */
+  bool update(const T &mapIn, T &mapOut) override;
+
+  /*!
+   * Averages (unweighted) [2*radius + 1] cells at a time along the rows of a gridmap layer
+   */
+  void HorizontalBoxBlur(const Eigen::MatrixXf &layerIn, Eigen::MatrixXf &layerOut, int r);
```
ConnorNeed: We don't really need to pass in r since HorizontalBoxBlur is a member function it can access radius_ directly

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+}
+
+template <typename T>
+void BoxBlurFilter<T>::VerticalBoxBlur(const Eigen::MatrixXf &layerIn, Eigen::MatrixXf &layerOut, int r)
```
ConnorNeed: Same comments apply as above

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+  HorizontalBoxBlur(mapIn[inputLayer_], mapOutHorz[outputLayer_], radius_); //Blurs input layer from mapIn horizontally, saves blurred values to mapOutHorz on outputLayer_
+  VerticalBoxBlur(mapOutHorz[outputLayer_], mapOut[outputLayer_], radius_); //Blurs outputLayer_ from mapOutHorz vertically, saves blurred values to mapOut on outputLayer_
+  //Box blur complete!
+  return true;
+}
+
+template <typename T>
+void BoxBlurFilter<T>::HorizontalBoxBlur(const Eigen::MatrixXf &layerIn, Eigen::MatrixXf &layerOut, int r) 
+{
+  int height = layerOut.rows();
```
ConnorNeed:
```suggestion
  const int height = layerOut.rows();
```

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+  int width = layerOut.cols();
```
ConnorNeed:
```suggestion
  const int width = layerOut.cols();
```

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  for (int y = 0; y < height; y++) {
+    for (int x = 0; x < width; x++) {
+      //ensuring that max and min cells in radius do not exceed dimensions of gridmap layer
+      int minCoeff = std::max(0, x - r);
```
ConnorNeed:
```suggestion
      const int minCoeff = std::max(0, x - r);
```

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
```diff
@@ -0,0 +1,145 @@
...
+      int maxCoeff = std::min(width - 1, x + r);
```
ConnorNeed:
```suggestion
      const int maxCoeff = std::min(width - 1, x + r);
```

ConnorNeed (commented):
Great work! Good use of header documentation. Your implemtation file could use less redundant comments.

omeletteee:
Thank you for the feedback! I've (hopefully) committed all the code changes you posted here and will now work on cleaning up my comments to be less redundant + add a counter for the number of valid values

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
omeletteee: Noted. I've removed the unnecessary comments. Thank you!

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
omeletteee: Removed comment!

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
omeletteee: Changed variable to numValues instead of numHoles! Very good point - it makes the math a lot cleaner too.

src/Nav/cprt_gridmap_filters/src/BoxBlurFilter.cpp
omeletteee: Removed unnecessary comments, thanks for pointing this out!

ConnorNeed (approved):
Great work Lauren. We will talk tonight about a new task!

PR #56, opened 2026-01-20
Integration chassis

ConnorNeed:
1. Adds "systemd-journal" to docker group. This maps to the "input" group on the host. Yes I hate this solution as well but it is fast and it works and means we can read the joystick without anything else necessary.
2. Adds Odrive ros2 package. Note: I did not add it as a git submodule since I had to make some changes to it. Just some extra logging for bad can bus (because we know we never have can issues) and the ability to scale the values to account for gear ratios. In the future we will want to read the rest of the telemetry as well and will need to add more code. Ignore everything under ros_odrive except where multiplier is used in src/HW-Devices/ros_odrive/odrive_ros2_control/src/odrive_hardware_interface.cpp since the rest is open source.
3. Remove 2 unused functions in src/Teleop-Control/swerve_controller/src/axle.cpp
4. Measured the rover for src/URDF/rover_urdf/config/ros2_controllers.yaml config
5. Configured src/URDF/rover_urdf/urdf/chassis_urdf.ros2_control.xacro for odrive
6. Fixed URDF so the wheels spin the correct direction in RViz

Rover was driving! (Until it broke but we don't talk about that)


Darren-Wallace (approved):
Love to see it driving 👍

PR #57, opened 2026-01-22
Arm + Drive modeless controll scheme configuration + switched controls to position mode

PriSey:
Created new modeless package to launch drive and arm on two controllers without switching
Switch arm modes to position

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ log/
 build*/
 *.vscode
 raman/
+src/kindr
 
 #Files
 *metadata.json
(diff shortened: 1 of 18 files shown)


ConnorNeed (changes requested):
The goal is to bring this new package up to minimal viable product in steps while keeping things simple.

The old code package was wildly complicated for doing such a simple task. Simpler is better. Unfortunately this brings back all the complexity of the old package.

Changing to position mode is outside the scope of this task and was already done by default in the newest urdf which has a different structure

My recommendation is start from scratch on a new branch right off of main. Create a new package then add two nodes.

The drive node subscribes to the drive joy topic and publishes a twist message.

The arm node subscribes to the arm joy topic and publishes a JointJog message.

The simpler the better.

Then add the launch file to launch those 2 nodes plus the two joy nodes with correct topic mapping and open a new pull request.

PR #58, opened 2026-01-24
Add yolov9 mallet model

ErikCald:
I trained a yolov9 model to detect the orange rubber mallet using google colab, converted it to a onnx model and set up the deepstream files to run it.

I also kept track of what I created to download roboflow datasets, re-split them, and other commands and google colab stuff. It is all documented in `colab_yolo_training_docs.md` so that someone else coming along in the future can have a head start.

https://jira.engsoc.net/browse/ROVER-667


ErikCald:
@ConnorNeed lint check if failing which I think you know. I made sure my new python files are passing it regardless.

MaybeWilli (approved):
Looks good to me, though I would like to mention that my exported .onnx file had output layer named "output" and not "output0", so if any issues arise, we'll probably just have to update the name (in mallet.txt)

src/Cameras/yolo_training/colab/requirements.txt
ConnorNeed, line 1: ?

ConnorNeed (approved):
Other than your pip requirements.txt being the full system reqs looks good to me. When you have a sec go through and double check the remaining copilot errors to either fix them or mark as resolved if they are hallucinations.

PR #59, opened 2026-01-24
Add --parallel-workers and --packages-select to build.sh

ErikCald:
Finally got around to improving the build.sh so that it has a --parallel-workers option and a --packages-select option.

Can be used like this:
```bash
./build.sh
./build.sh --parallel-workers 2 # up to how many CPUs you have (check by running $ nproc)
./build.sh --packages-select navigation
./build.sh --packages-select navigation localization gps --parallel-workers 1
```
I intentionally used the same names as the colcon build arguments so that memorizing these arguments can translate over to the colcon build arguments. Buuuutttt I also just commited a -s for --packages-select and a -w for --parallel-workers because I dont like typing things out lol

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,37 @@
 #!/bin/bash
 
+# Ensure script is run from the workspace root
+if [ ! -d "src" ]; then
+    echo "Error: 'src' directory not found. This script must be run from the workspace root."
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


ConnorNeed (approved):
This has been always something I wanted to add but never had. Thanks.

PR #62, opened 2026-01-27
feat: use onboard encoder + trapozoid trajectory

ConnorNeed:
Trap trajectory control from the odrive library is much easier on the motors. Ideally this is something that should be configureable but I am just going to make a Jira ticket for it and put it on the backlog. New members are always looking for easy non time sensitive tasks and this fits the bill.

Using a 60:1 scalling on the motors based on our gearbox(s). Also 2 of the signs needed to be flipped because of the direction of mounting.


Darren-Wallace (approved):
Looks good 👍

PR #63, opened 2026-01-29
[DRAFT] Manual mode end effector control

Jayden442:
This is for task [ROVER-681](https://jira.engsoc.net/browse/ROVER-681)

- Added end effector joint in arm_urdf.ros2_control.xacro (has placeholder CAN ID)
- In arm.cpp and arm.hpp, added parameters for claw open and close
- Arm will move towards closed position when close button is held, move towards open position when open button is held (Previously would just move to a position)

### DIFF
--- a/src/Teleop-Control/joystick_control/include/arm.hpp
+++ b/src/Teleop-Control/joystick_control/include/arm.hpp
@@ -35,6 +35,8 @@ private:
   int kAct1Axis;
   int kAct2Axis;
   int kElbowYaw;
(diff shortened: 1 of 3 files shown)


ConnorNeed:
This makes a lot of assumptions which are false. 

Joint_7 is not in the urdf, moveit has no mechanism of doing anything with it.

There is no encoder on that motor so there is no mechanism to do position control.

Let's close this one down and regroup tonight

PR #65, opened 2026-02-03
Create aruco localizer node for indoor localization loop closure

ErikCald:
An aruco board is a collection of aruco tags that are fixed relative to each other. I designed this to always use pairs of aruco tags as a aruco board to guarantee consistency. 

This PR does:
- A node to allow the zed camera to generate a config of a specific aruco board (accurately design where the tags are relative to each other)
- A node to read zed camera images from the zed wrapper and publish where aruco boards are located in 3D
- A node to take the aruco board locations and be a localizer that uses the aruco boards as fixed points (the aruco boards cannot move or the node needs to be restarted for it to learn the new aruco board location)

This should be a far more consistent and trustworthy setup than slam but does require the zed camera to have aruco boards in view to localize and do "loop closure". I only have 1 board right now I need to make more... 

I also have a zed svo2 recording which has an aruco board in the video and this works great to have good localization in recordings too, which is crazy useful for testing with recordings. But it needs to ALWAYS have a board in view because it wont have wheel odom when replaying the recording.


ConnorNeed (approved):
Blind approval. If you say it works I am happy

PR #68, opened 2026-02-08
Fix: Framerate control

jujuz455:
**Caps**:corrected to video/x-raw(memory:NVMM)


ConnorNeed (approved):
Looks good to me. As soon as you have tested it you are good to merge

jujuz455:
IMG_0099
I think this means success? @ConnorNeed

ConnorNeed:
>  I think this means success? @ConnorNeed

Set the framerate to 1 and watch the little static square in the bottom corner. Make sure it looks like it is only changing once per second. Then set the framerate to 30 and make sure it goes back to full speed.

jujuz455:
> Set the framerate to 1 and watch the little static square in the bottom corner. Make sure it looks like it is only changing once per second. Then set the framerate to 30 and make sure it goes back to full speed.

I will test it again tomorrow

PR #69, opened 2026-02-09
Canal test

ConnorNeed:
All my fixes that I needed for the canal test

- Adds a robo_claw node. Right now only does simple percent output control
- Adds dialout group to dev container to allow the dev container to connect to the robo_claw
- Adds all base station nodes to the launch file
- Fixed device ID for gps module

I wouldn't worry about the robo-claw node too much. The library is taken from the official repo and I don't expect the node to be used for very long

### DIFF
--- a/src/Cameras/video_streaming/config/cameras.yaml
+++ b/src/Cameras/video_streaming/config/cameras.yaml
@@ -8,7 +8,7 @@ input_node:
     - "Bottom"
     - "Test"
     Drive:
-      path: "/dev/v4l/by-id/usb-Arducam_Arducam_B0495__USB3_2.3MP_-video-index0"
+      path: "/dev/v4l/by-id/drive_camera"
       type: 0
     EndEffector:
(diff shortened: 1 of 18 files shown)


src/Cameras/video_streaming/config/cameras.yaml
```diff
@@ -8,7 +8,7 @@ input_node:
     - "Bottom"
     - "Test"
     Drive:
-      path: "/dev/v4l/by-id/usb-Arducam_Arducam_B0495__USB3_2.3MP_-video-index0"
```
jujuz455, line 11: Does this change affect the SRT node I'm working on?

ConnorNeed: Yes but also no. It's just the path id for the camera. The default sym link for that camera sometimes pops up differently so sometimes it works and sometimes it doesn't. So I added a udev rule so it always pops up as /dev/v4l/by-id/drive_camera making it more consistent.

Other than more consistently connecting to that specific camera (when plugged in) it doesn't change what you are working on at all

jujuz455 (approved):

PR #70, opened 2026-02-09
Fix bug in rolling costmap setup

ErikCald:
I had an issue where both costmaps were flickering out for like 0.2 seconds every second on rviz. It also saw that the planner was planner through obstacles when this happening meaning what I saw was truly the costmap plugin show a fully unknown costmap. When I turn off rolling costmap, the issue goes away. So I'm pretty confident this issue is because the rolling window config is clearing the map and so we need the gridmaplayer plugin to track the costmap state and handle remembering obstacles. 

@ConnorNeed  can you look through this code and let me know if it makes sense with your knowledge of the costmap plugin system?


ConnorNeed (approved):
Makes complete sense to me. 

The only thing I will say is we might still want to guard against writing to master until we have recieved at least one map. That's how it looks like the [other](https://github.com/ros-navigation/navigation2/blob/main/nav2_costmap_2d/plugins/static_layer.cpp#L164) plugins do it. In terms of the second AI comment I am okay with throwing and crashing the system with a very detailed error rather than trying to continue without our main vision mapping.

The build machine might not be plugged into the router. I'll plug it back in when I get there tonight but feel free to merge in when you are ready.

src/Nav/cprt_costmap_plugins/src/gridmap_layer.cpp
```diff
@@ -170,50 +170,53 @@ bool GridmapLayer::getTransform(
...
-
-  // Iterate through the grid map and copy the values to the costmap
-  for (grid_map::GridMapIterator it(gridmap_in_); !it.isPastEnd(); ++it) {
-    const grid_map::Index index(*it);
-    const float value = gridmap_in_.at(layer_name_, index);
-    const auto cost = interpretValue(value);
-    if (cost == NO_INFORMATION) {
-      continue;
+  // When new grid map data has arrived, copy it into the internal buffer
+  if (has_updated_data_) {
```
ConnorNeed:
```suggestion
  if (has_updated_data_ ||  layered_costmap_->isRolling()) {
```

How much does this hurt performance?

When costmap is located at (x1, y1) then translated to the costmap frame and copied over to the intenal layer. Then the next iteration happens and no new message has come in. We don't copy over to the intenal layer since no new map has happened. Internal layer is still based on (x1, y1) even though the robot has moved to (x2, y2). We don't move fast enough to really notice it but if we stop recieving data (very possible if the zed crashes) then the rolling costmap will stay the same even if the rover moves for a long time.

ConnorNeed (commented):
Sorry thought of this after I approved the first time

PR #80, opened 2026-03-23
Talon Load encoder (quadrature control, pwm load)

ConnorNeed:
Basic testing on Saturday was fine. Didn't try to kill itself under limited testing. Still stay cautious. Have a dedicated e-stop person.


ConnorNeed:
With last night's testing I am confident that this is at least better than what is currently in main. I know Darren is busy with capstone so I will YOLO merge to avoid get the arm operational as soon as possible

PR #83, opened 2026-04-04
chore: pylint in ci/cd + pylint fixes

ConnorNeed:
Caught some bugs using pylint:
1. Removed pi servo controller and i2c servo controller because they were broken since migrating our servo package over to the new API
2. Python2 vs Python3 mismatches
3. Import errors
4. Unused imports

Also lints and formats inside the docker build process for MRs (Not for the official builds)


Landwhich (approved):
LGTM

PR #85, opened 2026-04-04
fix: rtp fec decoding

ConnorNeed:
I did not realize that the fec decoder requires to be used with rtpstorage and do-lost=1 being set on rtpjitterbuffer.


jujuz455 (approved):
I didn't know the FEC decoder was that strict about the jitterbuffer config. Does it affact the latency directly?

PR #86, opened 2026-04-07
Grip control

jujuz455:
Added ControlMode::Current to the passthrough conditional block in phoenix_node.hpp.

### DIFF
--- a/src/HW-Devices/ros_phoenix_humble/include/ros_phoenix/phoenix_node.hpp
+++ b/src/HW-Devices/ros_phoenix_humble/include/ros_phoenix/phoenix_node.hpp
@@ -7,10 +7,10 @@
 #include "ros_phoenix/base_node.hpp"
 
 #define Phoenix_No_WPI // remove WPI dependencies
+#include "ctre/Phoenix.h"
 #include <chrono>
 #include <stdexcept>
-
-#include "ctre/Phoenix.h"
+#include <type_traits>
 
 using namespace rclcpp;
 using namespace std::chrono_literals;
@@ -68,17 +68,28 @@ public:
       // CTRE library expects velocity in units/100ms
       this->controller_->Set(mode, control_msg->value / 10.0 /
                                        this->sensor_multiplier_);
-    } else if (mode == ControlMode::Position) {
+      return;
+    }
+    if (mode == ControlMode::Position) {
       this->controller_->Set(mode, (this->sensor_offset_ + control_msg->value) /
                                        this->sensor_multiplier_);
-    } else if (mode == ControlMode::PercentOutput ||
-               mode == ControlMode::Disabled) {
-      this->controller_->Set(mode, control_msg->value);
-    } else {
-      this->controller_->Set(ControlMode::Disabled, 0.0);
-      RCLCPP_WARN(this->get_logger(), "Invalid control mode: %d",
-                  static_cast<int>(mode));
+      return;
     }
+    if (mode == ControlMode::PercentOutput || mode == ControlMode::Disabled) {
+      this->controller_->Set(mode, control_msg->value);
+      return;
+    }
+    if constexpr (!std::is_same_v<
+                      MotorController,
+                      ctre::phoenix::motorcontrol::can::VictorSPX>) {
+      if (mode == ControlMode::Current) {
+        this->controller_->Set(mode, control_msg->value);
+        return;
+      }
+    }
+    this->controller_->Set(ControlMode::Disabled, 0.0);
+    RCLCPP_WARN(this->get_logger(), "Invalid control mode: %d",
+                static_cast<int>(mode));
   }
 
   virtual rcl_interfaces::msg::SetParametersResult


src/HW-Devices/ros_phoenix_humble/include/ros_phoenix/phoenix_node.hpp
```diff
@@ -74,6 +74,8 @@ class PhoenixNode : public BaseNode {
     } else if (mode == ControlMode::PercentOutput ||
                mode == ControlMode::Disabled) {
       this->controller_->Set(mode, control_msg->value);
+    } else if (mode == ControlMode::Current) {
+      this->controller_->Set(mode, control_msg->value);
```
ConnorNeed: Given that the executed code is exactly the same, you can do something like this:
```suggestion
} else if (mode == ControlMode::PercentOutput ||
           mode == ControlMode::Current ||
           mode == ControlMode::Disabled) {
      this->controller_->Set(mode, control_msg->value);
```

src/Arm/arm_control/launch/end_effector.launch.py
```diff
@@ -21,6 +21,10 @@ def generate_launch_description():
                     {"max_voltage": 24.0},
                     {"max_current": 6.0},
                     {"brake_mode": True},
+                    {"P": 0.0},
+                    {"I": 0.002},
+                    {"D": 0.0},
+                    {"F": 0.3},
```
ConnorNeed: These are very interesting PID values. Any particular reason for them?

src/Teleop-Control/joystick_control/src/arm.cpp
```diff
@@ -19,19 +21,37 @@ arm::arm() : Node("arm_node") {
 
   RCLCPP_INFO(this->get_logger(), "Arm controller started");
 }
+
 void arm::endeffector_control(
     std::shared_ptr<sensor_msgs::msg::Joy> joystickMsg) {
-  auto &buttons = joystickMsg->buttons;
+  auto &axes = joystickMsg->axes;
```
ConnorNeed: What axis are you suggesting we use? All of them are in use for controlling the arm.

src/Teleop-Control/joystick_control/src/arm.cpp
```diff
@@ -19,19 +21,37 @@ arm::arm() : Node("arm_node") {
...
+    if (grip_state_ == GRIP_CLOSING || grip_state_ == GRIP_HOLDING) {
+      grip_state_ = GRIP_HOLDING;
+      eef_control_msg.mode = ros_phoenix::msg::MotorControl::CURRENT;
+      eef_control_msg.value = -std::abs(kHoldCurrent); 
+    } else {
+      grip_state_ = GRIP_IDLE;
+      eef_control_msg.mode = ros_phoenix::msg::MotorControl::DISABLED;
+      eef_control_msg.value = 0.0;
+    }
   }
```
ConnorNeed: Not sure I follow the logic here.

ConnorNeed (commented):
I'll let Darren comment on the human control side more

ConnorNeed:
CI/CD pipeline found an issue with building it. Looks like a templating bug since not every motor controller supports current control. A few ways to fix it but I will let you research since templates in C++ are a really cool mechanism. Try to fix it using a template overload but if you can't I am okay with just removing SPX support (We don't use it anyways).

https://api.ctr-electronics.com/phoenix/stable/cpp/_control_mode_8h.html


You might find: if constexpr (std::is_same_v<X,Y>) {} helpful

jujuz455:
> CI/CD pipeline found an issue with building it. Looks like a templating bug since not every motor controller supports current control. A few ways to fix it but I will let you research since templates in C++ are a really cool mechanism. Try to fix it using a template overload but if you can't I am okay with just removing SPX support (We don't use it anyways).
> 
> https://api.ctr-electronics.com/phoenix/stable/cpp/_control_mode_8h.html
> 
> You might find: if constexpr (std::is_same_v<X,Y>) {} helpful

Thanks, I will try and let you know :D

src/HW-Devices/ros_phoenix_humble/include/ros_phoenix/phoenix_node.hpp
```diff
@@ -74,6 +74,14 @@ class PhoenixNode : public BaseNode {
     } else if (mode == ControlMode::PercentOutput ||
                mode == ControlMode::Disabled) {
       this->controller_->Set(mode, control_msg->value);
+    } else if (mode == ControlMode::Current){
```
ConnorNeed: I have a feeling this is still going to fail to compile. Any reference to ControlMode::Current when ControlMode is templated as ctre::phoenix::motorcontrol::VictorSPXControlMode will not compile. 

the if constexpr allows the compiler to evaluate at compile time and ignore inside the brackets if false. All references to ControlMode::Current must be guarded by the is_same_v.

src/HW-Devices/ros_phoenix_humble/include/ros_phoenix/phoenix_node.hpp
```diff
@@ -64,21 +64,32 @@ class PhoenixNode : public BaseNode {
     BaseNode::set(control_msg);
 
     ControlMode mode = static_cast<ControlMode>(control_msg->mode);
+    // CTRE library expects velocity in units/100ms
     if (mode == ControlMode::Velocity) {
-      // CTRE library expects velocity in units/100ms
```
ConnorNeed, line 68: I would leave this comment where it was

ConnorNeed (approved):
Fix the comment, wait for pass then you are good to squash merge

PR #88, opened 2026-04-09
Modeless control

opom1231:
Implemented modeless control for the arm and drive systems, including MoveGroup action clients for presets. Integrated the latest science node and resolved CMake conflicts.


ConnorNeed:
Thank you for the starting point. We are going to use some of this going forward.

PR #91, opened 2026-04-12
feat: PTZ camera control

ConnorNeed:
The entire servo_pkg is a mess and could use some massive clean up. In the mean time this does what we need to control the gimble camera from the PS controller.


codeflight1 (approved):
LGTM

PR #93, opened 2026-04-12
Cleanup

ConnorNeed:


ConnorNeed:
Removed some unused packages to avoid bloat and confusion. If we end up needing any of this back, that's what git is for I suppose.

Added a script to auto-generate our rosdep keys file. With the removed packages it seems like 9 rosdeps are no longer needed. Hopefully that will make our images a little smaller (Not much but a drop in the bucket). Post URC I might try to optimize docker image size.

Relying on the cmake, gcc and pylint to make sure we have the dependacies we need. Another reason I wanted to do this now since it get's riskier the closer we are to comp.

codeflight1 (approved):
LGTM. Thanks for cleaning it up!

PR #100, opened 2026-05-04
Add keyboard pose estimation using ArUco markers

jujuz455:
- Implemented KeyboardPnPLocator node for estimating keyboard pose using per-marker PnP pose estimation with ArUco markers.
- Added camera calibration and keyboard board configuration YAML files.
- Created launch file for the keyboard PnP locator node.
- Generated keyboard key layout for the Redragon K552 ANSI TKL keyboard.
- Updated setup.py to include new node and configuration files.
- Enhanced video streaming node to publish images from the arm camera.
- Integrated GStreamer appsink for capturing frames in the video streaming pipeline.
IMG_1533


ConnorNeed (commented):
Looks good to me. I don't think we are going to have an arm good enough to use it for typing but maybe we could use it for Nav? @ErikCald 

Either way good job and I am sure this will come in handy next year.

PR #101, opened 2026-05-05
Send compressed gridmap and costmaps over network

ErikCald:
The costmaps and gridmaps should be able to compress down, especially because alot of the costmap is uniformly unknown. o I made some nodes that will compress them and send as uint8 arrays over the antennas to hopefully make it lighter on the system and make the data more available for me.

Draft PR because i dont know when I will get a chance to actually test if this works and how much it compresses it down by. Need to test it IRL. I could do it on my jetson orin nano probably.


ConnorNeed:
My main concern is how heavy the processing for this would be. Python subscribers are notoriously bad which large data. We can test this weekend :)

ErikCald:
> My main concern is how heavy the processing for this would be. Python subscribers are notoriously bad which large data. We can test this weekend :)

Yeah i probably need to implement in cpp, which also means it can be composable and do shared memory zero copy in order to get the costmaps and gridmap. But this will still be a good proof of concept to see how much compression actually helps or not

ErikCald:
```bash
Compressed full costmap: 250104 bytes -> 725 bytes (99.7% smaller, 0.00x of original)
```

```bash
Compressed gridmap: 4889 bytes -> 1097 bytes (77.6% smaller, 0.22x of original)
Compressed send rate: 0.044 Mbps over last 10s (54999 bytes, avg reduction 77.5%, avg ratio 0.22)
```

src/Nav/compressed_telemetry/compressed_telemetry/gridmap_decompressor.py
```diff
@@ -0,0 +1,159 @@
...
+            grid_msg.info.pose.orientation.w = qw
+
+            grid_msg.layers = ["elevation", "traversability"]
+            grid_msg.basic_layers = []  # Usually empty for basic visualization
+            grid_msg.outer_start_index = outer_start
+            grid_msg.inner_start_index = inner_start
+
+            # Package arrays into Float32MultiArray wrappers
+            elev_multi = Float32MultiArray()
+            elev_multi.data = elev_float.tolist()
```
ErikCald: Gemini comment:

The problem comes down to missing MultiArrayLayout metadata in the Float32MultiArray objects you are publishing, which crashes the C++ subscriber (RViz) and subsequently locks up the Python publisher.

Here is the exact breakdown of what is happening:

1. Why RViz crashes (Exit Code 139 / Segfault)
Inside the GridMap message structure, each layer (elevation, traversability) is repesented by a Float32MultiArray. The C++ codebase for grid_map_ros (and the RViz plugin by extension) uses an internal helper function called isRowMajor that checks how the array is structured. It specifically checks message.layout.dim[0].label to see if the first dimension is named "column_index" or "row_index".
Because your script leaves layout completely empty, dim is an empty array. When C++ tries to access index [0] of that empty array, it results in an out-of-bounds memory access (Segmentation fault), instantly killing RViz.

2. Why the Python node hangs
When a subscriber (RViz) dies abruptly (segfault) right as the Python node is busy serializing and transmitting a massive GridMap message over the network layer (DDS), the underlying DDS middleware can sometimes get stuck waiting for acknowledgments or holding locks in the publisher. Additionally, calling .tolist() on massive NumPy arrays causes a CPU spike that locks up the Python GIL, making the node feel unresponsive to Ctrl+C.

The Fix:
You need to explicitly construct and attach a MultiArrayLayout to both elev_multi and trav_multi before publishing. You have to specify the two dimensions (MultiArrayDimension) based on the grid shape (length_x / resolution and length_y / resolution) and label them as "column_index" and "row_index".

ErikCald:
@ConnorNeed 
Ported to cpp and tested on my jetson orin nano with a .svo2 file from mission control.
```bash
Costmaps: Avg compression over 20 s: 99.6% (avg 250104 bytes -> 1006 bytes), 
         0.014 Mbps, incoming 1.80 Hz, compressed 1.80 Hz
```
```bash
Gridmaps: Avg compression over 20 s: 31.8% (avg 4889 bytes -> 3332 bytes), 
         0.077 Mbps, incoming 2.90 Hz, compressed 2.90 Hz
```

My jetson orin nano can only publish the gridmaps at 2.9 hz because its a weak cpu. Im confident this will work fine at 18hz on the jetson agx orin too.

ConnorNeed (approved):
LGTM

PR #105, opened 2026-05-08
Arm presets

ConnorNeed:
Adds arm presets to the SRDF and hooks to move the arm to those presets from the WebUI. To add new ones all you have to do is add them to the SRDF and the webUI will automatically include them in the dropdown


Darren-Wallace (approved):
Looks good to me 👍

PR #106, opened 2026-05-08
Add PDB_Rails node

codeflight1:
Reports power good signals read from the SmartPDB and commands the toggle pins to enable and disable power rails.


ConnorNeed (approved):
Awesome stuff!
Make sure to use squash merge to keep the git history clean

PR #108, opened 2026-05-09
fix: only allow preset moves in when joystick disabled

ConnorNeed:
Add a check to see the control type. Only move in joystick disabled mode.


Darren-Wallace (approved):
LGTM 👍

PR #109, opened 2026-05-09
Fixes may9

ConnorNeed:
Moves more nodes to the rover side to reduce the amount of web traffic during discovery.
Practically this means that the webUI top field needs to be set to the rover IP instead of the base station.


Darren-Wallace (approved):
LGTM 👍

PR #110, opened 2026-05-09
Science controls updated

TomasWilliston:
Implements a new ROS node for communicating with the science module, and updated drill controls (for roboclaw, needs to be changed to fit the talons instead)

The ROS node creates a topic to send a command to create a PWM signal on the science module's ESP32, and a topic that publishes the sensor readings from the module.


ConnorNeed (approved):
Looks good to me. Awesome work. 

I fixed the format issue so we are good to merge in. We will update the launch files, webUI and and switch to the talons tomorrow.

PR #117, opened 2026-05-15
Headers esp code

ConnorNeed:
Adds headers and switches libraries to avoid the crash


TomasWilliston (approved):
Looks good, thanks for taking care of that.

PR #129, opened 2026-05-20
fix: panoramic

ConnorNeed:
Left in some of our logging


src/HW-Devices/science_sensors/science_sensors/panoramic.py
```diff
@@ -135,15 +149,21 @@ def done_callback(future):
         result = future.result()
         images = []
         for i in range(self.num_images):
-            pan_angle = i * (6.28 / self.num_images)
+            pan_angle = i * (3.0 / self.num_images)
```
codeflight1: Are we leaving it as only half or do we need to do 360 for competition?

codeflight1 (approved):

PR #133, opened 2026-05-26
Science updates

codeflight1:
- new servo control
 - polarimeter support

I had to downgrade scipy and haven't had time to build the full image so this *should* work but needs to be checked (also anywhere else we use the scipy pip package). The polarimeter node now produces labeled graphs, webui pr to follow.

Additionally, the esp code changes have not been tested. I hope my dummy reads will fix the problem but if not we need to be ready to think of something else.

### DIFF
--- a/src/HW-Devices/science_sensors/science_sensors/panoramic.py
+++ b/src/HW-Devices/science_sensors/science_sensors/panoramic.py
@@ -149,7 +149,7 @@ class PanoramicNode(Node):
         result = future.result()
         images = []
         for i in range(self.num_images):
-            pan_angle = i * (3.0 / self.num_images)
+            pan_angle = i * (6.28 / self.num_images)
             self.get_logger().info(f"Moving servo to {pan_angle} radians")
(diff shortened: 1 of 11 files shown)


src/HW-Devices/science_sensors/science_sensors/panoramic.py
```diff
@@ -149,7 +149,7 @@ def done_callback(future):
         result = future.result()
         images = []
         for i in range(self.num_images):
-            pan_angle = i * (6.28 / self.num_images)
+            pan_angle = i * (3.0 / self.num_images)
```
ConnorNeed: I have tested this with 6.28 and it works well. Can we leave it at 6.28?

ConnorNeed (commented):
We will test tomorrow

src/HW-Devices/science_sensors/science_sensors/panoramic.py
codeflight1: Oops, that was just me goofing the merge. Fixed

PR #154, opened 2026-07-01
Arm API update

ConnorNeed:
Still needs some more tuning but this is making the preset controls much nicer and more reliable.

Also fixed an emulation build failure (I think)


codeflight1 (dismissed):
Nice work, I like the new API. Would you mind adding a publisher for the arm control mode? I'm also not seeing the `MoveGroupStatus` message being read anywhere, maybe we should throw that on the webUI as well.

PR #155, opened 2026-07-04
Getting a Proper Heartbeat Error Reading

omeletteee:
Small fix, can test it with the rover and if it doesn't work it can be reverted / changed.


src/HW-Devices/ros_odrive/odrive_ros2_control/src/odrive_hardware_interface.cpp
codeflight1:
```suggestion
    status_msg.active_errors = axis.axis_error_
```

codeflight1 (changes requested):

ConnorNeed:
Rebased and corrected the boolean logic @codeflight1 and @omeletteee

codeflight1 (approved):

PR #159, opened 2026-07-06
Video stream mosaic

jujuz455:
This PR adds a second independent video output that streams a mosaic view of all connected cameras via WebRTC, separate from the existing operator view.

**Dual compositor pipeline**: Each camera source is now split via tee and fed into two compositors in parallel : 
the existing compositor (operator view, dynamically controlled via video_cb) and a new mosaic_compositor (fixed grid layout)

**Automatic grid layout**: Added set_mosaic_layout(), which computes a near-square grid (columns/rows) based on the number of active cameras and sets each pad's xpos, ypos, width, and height accordingly

**New MosaicWebRtcNode:** Reads the mosaic feed via interpipesrc, encodes it with nvv4l2h264enc using all-intra encoding (iframeinterval=1) so frame drops don't corrupt the picture (similar robustness to MJPEG, but better compression), and streams it out via webrtcsink with its own signalling server (default port 8444)

### DIFF
--- a/src/Cameras/video_streaming/CMakeLists.txt
+++ b/src/Cameras/video_streaming/CMakeLists.txt
@@ -28,6 +28,7 @@ add_library(${PROJECT_NAME} SHARED
   src/input_node.cpp
   src/detect_node.cpp
   src/webrtc_node.cpp
+  src/mosaic_webrtc_node.cpp
   src/srt_node.cpp
   src/srt_node_client.cpp
   src/rtp_node.cpp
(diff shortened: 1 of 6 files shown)


ConnorNeed:
Didn't do full setup over our regular comms or test that many cameras but it works well enough to merge in so it will be ready for Saturday night. Awesome job

I also rebased off latest main changes.

ConnorNeed (approved):
Awesome job.

Take a look at my very few bug fixes to make sure you understand them (ask any questions as well) and then go ahead and squash merge!

PR #164, opened 2026-07-09
Morse gstreamer

TomasWilliston:
Adds a gstreamer plugin (+ helper modules) for parsing morse code from a red blinking LED, as well as another plugin + python script for testing the detection without a camera.


src/Cameras/video_streaming/gst-plugins/morseLED/gstmorse_led.cpp
```diff
@@ -0,0 +1,531 @@
...
+                                       gboolean led_on) {
+  const gint frame_w = GST_VIDEO_FRAME_WIDTH(frame);
+  const gint frame_h = GST_VIDEO_FRAME_HEIGHT(frame);
+  const gint x0 = std::max(0, std::min(roi_x, frame_w - 1));
+  const gint y0 = std::max(0, std::min(roi_y, frame_h - 1));
+  const gint x1 = std::min(frame_w - 1, x0 + static_cast<gint>(roi_w) - 1);
+  const gint y1 = std::min(frame_h - 1, y0 + static_cast<gint>(roi_h) - 1);
+  if (x1 <= x0 || y1 <= y0) {
+    return;
+  }
```
ConnorNeed, line 661: I think this guarding logic could be a little nicer:

If you enforce that roi_x and roi_y are always >=0 on setup you can skip checking for that here.

```suggestion
  const gint frame_w = GST_VIDEO_FRAME_WIDTH(frame);
  const gint frame_h = GST_VIDEO_FRAME_HEIGHT(frame);
  if (frame_w <=0 || frame_h <=0) {
    // Log as a warning
    return;
  }
  
  const gint x0 = std::min(roi_x, frame_w - 1);
  const gint y0 = std::min(roi_y, frame_h - 1);
  const gint x1 = std::min(x0 + static_cast<gint>(roi_w), frame_w - 1);
  const gint y1 = std::min(y0 + static_cast<gint>(roi_h), frame_h - 1);
  if (x1 <= x0 || y1 <= y0) {
    // Log as a warning
    return;
  }
```

Gets rid of the nested min-max and allows you to debug the difference between the 2 errors making things easier to debug.

src/Cameras/video_streaming/gst-plugins/morseLED/gstmorse_led.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+  gst_morse_led_finish_calibration(self, frame);
+  self->calibrate = FALSE;
+  g_object_notify(G_OBJECT(self), "calibrate");
+  gst_morse_led_reset_calibration(self);
+}
+
+static void gst_morse_led_set_pixel(guint8 *data, gint stride, gint x, gint y,
```
ConnorNeed: This gets called a tonne - should inline it

```suggestion
static inline void gst_morse_led_set_pixel(guint8 *data, gint stride, gint x, gint y,
```

src/Cameras/video_streaming/gst-plugins/morseLED/gstmorse_led.cpp
```diff
@@ -0,0 +1,531 @@
...
+  const guint8 b = 0;
+  constexpr guint kThickness = 2;
+
+  for (guint t = 0; t < kThickness; ++t) {
+    const gint top = std::min(y1, y0 + static_cast<gint>(t));
+    const gint bottom = std::max(y0, y1 - static_cast<gint>(t));
+    const gint left = std::min(x1, x0 + static_cast<gint>(t));
+    const gint right = std::max(x0, x1 - static_cast<gint>(t));
+
+    for (gint x = x0; x <= x1; ++x) {
```
ConnorNeed:
```suggestion
    for (gint x = left; x <= right; ++x) {
```

Very minor overdraw

src/Cameras/video_streaming/gst-plugins/morseLED/gstmorse_led.cpp
```diff
@@ -0,0 +1,531 @@
...
+      gst_morse_led_set_pixel(data, stride, x, top, r, g, b);
+      gst_morse_led_set_pixel(data, stride, x, bottom, r, g, b);
+    }
+    for (gint y = y0; y <= y1; ++y) {
```
ConnorNeed: Same here

src/Cameras/video_streaming/gst-plugins/morseLED/gstmorse_led.cpp
```diff
@@ -0,0 +1,531 @@
...
+static GstFlowReturn gst_morse_led_transform_frame_ip(GstVideoFilter *filter,
+                                                      GstVideoFrame *frame) {
+  GstMorseLED *self = GST_MORSE_LED(filter);
+
+  const GstClockTime pts = GST_BUFFER_PTS(frame->buffer);
+  gdouble timestamp_sec = 0.0;
+  if (GST_CLOCK_TIME_IS_VALID(pts)) {
+    timestamp_sec =
+        static_cast<gdouble>(pts) / static_cast<gdouble>(GST_SECOND);
+  }
```
ConnorNeed, line 710: Don't fail silently here. If pts is invalid then timestamp_sec remains 0.0 and the timing will probably not work. Log the issue as a Error and return early since things aren't going to go well anyways.

src/Cameras/video_streaming/gst-plugins/morseLED/gstmorse_led.cpp
```diff
@@ -0,0 +1,531 @@
...
+  gdouble calibration_start_sec;
+  gboolean calibration_started;
+  gdouble calibration_sum_x;
+  gdouble calibration_sum_y;
+  gdouble calibration_sum_weight;
+
+  MorseDecoder *decoder;
+  gboolean draw_roi;
+  gboolean last_led_on;
+};
```
ConnorNeed, line 82: Need a mutex for any parameters that can be changed at runtime such as roi_* values. Any parameters that can not be edited at runtime should be explicity set in their g_object_class_install_property call.

src/Cameras/video_streaming/gst-plugins/morseLED/gstmorse_led.cpp
```diff
@@ -0,0 +1,531 @@
...
+      self->calibration_sum_y += static_cast<gdouble>(y) * dominance;
+      self->calibration_sum_weight += dominance;
+    }
+  }
+}
+
+static void gst_morse_led_finish_calibration(GstMorseLED *self,
+                                             const GstVideoFrame *frame) {
+  if (self->calibration_sum_weight <= 0.0) {
+    return;
```
ConnorNeed, line 587: Log this warning

src/Cameras/video_streaming/gst-plugins/morseLED/morse_decoder.cpp
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  // classify if there's a transition in the signal.
+  if (next_led_on != decoder->led_on) {
+    const gdouble segment_duration =
+        timestamp_sec - decoder->last_transition_sec;
+    if (segment_duration < kMinTransitionUnits * decoder->unit_sec) {
+      decoder->last_timestamp_sec = timestamp_sec;
+      return;
+    }
```
ConnorNeed: This debouncing logic only does half the job.

Heres a case it fails:
LED has been ON for almost a dot
one bad reading dips below threshold_off
segment_duration is valid/large enough
decoder accepts OFF transition

Then a dash would get quantified as 2 dots.

A better mechanism is:
pending_state and pending_since

If a potential transition event happens then set pending_state and pending_since values before returning.

if (pending_state == last_state && last_state != new_reading)
  set pending values and return

if (last_state != new_reading && pending_state == new_reading)
  if it's been pending long enough
     transition
  else
     return

set pending state to current state

You will have to flush it out a little more but it shouldn't be too bad

src/Cameras/video_streaming/gst-plugins/morseLED/morse_decoder.cpp
```diff
@@ -0,0 +1,263 @@
...
+    decoder->symbol_buf[decoder->symbol_len] = '\0';
+  }
+}
+
+void classify_off_duration(MorseDecoder *decoder, gdouble off_duration) {
+  const gdouble letter_gap_threshold =
+      MORSE_LETTER_GAP_UNITS * kGapDetectRatio * decoder->unit_sec;
+  const gdouble word_gap_threshold =
+      MORSE_WORD_GAP_UNITS * kGapDetectRatio * decoder->unit_sec;
+  if (decoder->symbol_pending_gap && off_duration >= letter_gap_threshold) {
```
ConnorNeed, line 100:
```suggestion
  if (off_duration >= letter_gap_threshold) {
```

If symbol_pending_gap is false for some reason but symbol_len > 0, a long word gap can append a space without finalizing the current symbol first. I don't see it happening but since finalize_symbol() already returns if symbol_len == 0 there is no real downside that I can see

src/Cameras/video_streaming/gst-plugins/morseLED/morse_decoder.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include <cstring>
+
+namespace {
+constexpr gdouble kMinUnitSeconds = 0.015;
+constexpr gdouble kMaxUnitSeconds = 1.2;
+constexpr gdouble kMinTransitionUnits = 0.08;
+constexpr gdouble kGapDetectRatio = 0.8;
+constexpr gdouble kDotMaxUnits = 1.8;
+constexpr gdouble kDashMinUnits = 2.2;
+constexpr gfloat kOnMargin = 0.015f;
```
ConnorNeed: Can we make this configurable? If we aren't getting readings in the field we will need some knob to turn to make it work

src/Cameras/video_streaming/gst-plugins/morseLED/morse_decoder.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "morse_timing.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <cstring>
+
+namespace {
+constexpr gdouble kMinUnitSeconds = 0.015;
+constexpr gdouble kMaxUnitSeconds = 1.2;
+constexpr gdouble kMinTransitionUnits = 0.08;
```
ConnorNeed: At 66.7ms dit length this corresponds to 5.336ms. Our fastest camera only gets 60fps of 16.67ms. We might have to ignore all debouncing and pray that with enough attempts we get a good enough picture.

ConnorNeed (commented):
Great work

src/Cameras/video_streaming/gst-plugins/morseLED/gstmorse_led.cpp
```diff
@@ -0,0 +1,755 @@
...
+  roi_width = self->roi_width;
+  roi_height = self->roi_height;
+  g_mutex_unlock(&self->lock);
+
+  if (start_detection) {
+    gst_morse_led_update_calibration(self, frame, timestamp_sec);
+
+    g_mutex_lock(&self->lock);
+    roi_x = self->roi_x;
+    roi_y = self->roi_y;
```
ConnorNeed, line 724: Are these values expected to be different from line 694?

ConnorNeed (commented):
I don't think you need all these locks.

The streaming thread is single threaded so any object not changed in the set_params does not need protection. For params that don't need to be changed at runtime you can set them as GST_PARAM_MUTABLE_READY and then gstreamer won't let them change while the pipeline is ready

ConnorNeed (approved):
Looks good.
Good luck with testing this weekend

ConnorNeed:
You can either squash merge or clean up your git history into a couple more organized commits.

codeflight1 (approved):

PR #170, opened 2026-07-15
Windows dev container GUI support + missing rover_urdf view.rviz

jaydizz0:
Tested the dev container setup on Windows per Connor's request in discord. The
linux-lite-gui display args don't port over directly — Windows needs a
different mount path.

## Why the Linux args don't work on Windows

`/tmp/.X11-unix` is a WSLg tmpfs inside the user's WSL distro. The Docker
daemon runs in the separate `docker-desktop` distro and resolves mount
sources against *its own* filesystem, where that path doesn't exist. Docker's
response to a missing mount source is to silently create an empty directory
and mount that — so the container gets `/tmp/.X11-unix` with no `X0` socket
and no error anywhere.

Verified side by side:

```
docker run --rm -v /tmp/.X11-unix:/tmp/.X11-unix <img> ls /tmp/.X11-unix
  -> empty

docker run --rm -v /run/desktop/mnt/host/wslg/.X11-unix:/tmp/.X11-unix <img> ls /tmp/.X11-unix
  -> X0
```

`/run/desktop/mnt/host/wslg` is where Docker Desktop exposes the host's WSLg
socket inside its VM — reachable by the daemon.

Also `${localEnv:DISPLAY}` resolves to `1` (not `:0`) from a Dev Containers
window, so it's hardcoded to `:0`. WSLg always serves `:0`.

`.devcontainer/windows-gui/devcontainer.json` is a copy of the existing
`windows` config plus those two runArgs. Device mounts are deliberately
omitted since those host paths don't exist on Windows.

## Second commit: missing view.rviz

Unrelated, happy to split into its own PR if preferred.

`rover_urdf/launch/launch.py` defaults `rviz_config_file` to
`share/rover_urdf/rviz/view.rviz`, but that file has never existed in the
repo. `setup.py` globs `rviz/*`, which returns an empty list and silently
installs nothing — so RViz gets a path to a nonexistent file and falls back
to its built-in defaults (Fixed Frame `map`, which nothing publishes without
Nav2, and no RobotModel display). `use_rviz:=True` has always opened to a
blank grid with a frame error.

Adds the config with Fixed Frame `base_link` and RobotModel on
`/robot_description`. `setup.py` already had the install rule.

## Testing

Windows 11 + WSL2 + Docker Desktop, RTX 5060. RViz reports
`OpenGl version: 4.5 (GLSL 4.5)` and renders the rover URDF at ~31fps with
joint_state_publisher_gui driving the arm. No external X server (no VcXsrv).

Opened from a fork — I don't have push access to CPRT/rover.

### DIFF
--- a/.devcontainer/windows-gui/devcontainer.json
+++ b/.devcontainer/windows-gui/devcontainer.json
@@ -1,0 +1,39 @@
+{
+  "name": "CPRT Dev Container (Windows Host + GUI)",
+  "image": "cprtsoftware/rover:dev-amd64",
+
+  // GUI support via WSLg. Unlike the Linux configs, the X11 socket cannot be
+  // mounted from /tmp/.X11-unix: that path is a WSLg tmpfs inside the user's
+  // WSL distro, but the Docker daemon resolves mount sources against its own
+  // filesystem (the docker-desktop distro), where it does not exist. Docker
+  // Desktop exposes the host's WSLg socket inside the VM at the path below.
+  // DISPLAY is hardcoded because WSLg always serves :0, and ${localEnv:DISPLAY}
+  // does not resolve to the WSL value from a Dev Containers window.
+  "runArgs": [
+    "--env", "DISPLAY=:0",
+    "--volume", "/run/desktop/mnt/host/wslg/.X11-unix:/tmp/.X11-unix:rw"
+  ],
+  "mounts": [
+  ],
+  "containerEnv": {
+    "SSH_AUTH_SOCK": "/ssh-agent",
+    "NVIDIA_VISIBLE_DEVICES": "all",
+    "NVIDIA_DRIVER_CAPABILITIES": "all"
+  },
+
+  "remoteUser": "vscode",
+  "containerUser": "vscode",
+  "remoteEnv": {
+    "USER": "vscode"
+  },
+
+  "customizations": {
+    "vscode": {
+      "settings": {
+        "terminal.integrated.shell.linux": "/bin/bash"
+      },
+      "extensions": [ "ms-vscode.cpptools", "ms-azuretools.vscode-docker" ]
+    }
+  },
+  "postCreateCommand": "echo 'source /opt/ros/humble/setup.bash' >> ~/.bashrc"
+}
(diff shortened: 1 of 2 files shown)


src/URDF/rover_urdf/rviz/view.rviz
codeflight1, line 1: We already have a `moveit.rviz` in the `arm_control` package, could you update the urdf launch file to point at that config instead of adding a new one?

.devcontainer/windows-gui/devcontainer.json
codeflight1, line 1: It looks like this was copied from an old version of the windows config, we now use ROS jazzy and the user is ubuntu.

codeflight1 (changes requested):

.devcontainer/windows-gui/devcontainer.json
jaydizz0: my local main was 40 commits behind, so I copied the pre-Jazzy windows config without realizing. Will rebase onto current main and redo this from the dev-lite / ubuntu / jazzy version.

src/URDF/rover_urdf/rviz/view.rviz
jaydizz0: Thanks! I tried reusing moveit.rviz, but it doesn't have a RobotModel display. It uses MotionPlanning, which needs move_group running to show the robot. Since rover_urdf's launch doesn't start move_group, use_rviz would still open blank.

.devcontainer/windows-gui/devcontainer.json
jaydizz0: Rebased onto current main and updated it to dev-lite / ubuntu / jazzy. Tested on Windows 11 + WSL2: builds on dev-lite, runs as ubuntu, and RViz renders on Jazzy.

jaydizz0:
Thanks for the review. Pushed an update for both:

Devcontainer: my local main was behind, so I'd copied the old windows config — rebased onto current main and updated it to dev-lite / ubuntu / jazzy to match.

view.rviz: replied on that thread. moveit.rviz won't quite work as a drop-in here (no RobotModel display, so use_rviz would still open blank), so I kept the new config for now — happy to split it into its own PR if you'd rather keep this one focused on the devcontainer.

Tested on Windows 11 + WSL2: builds on dev-lite, runs as ubuntu, and RViz renders on Jazzy. Ready for another look whenever — thanks!

codeflight1 (approved):
Thanks for getting this supported!

PR #172, opened 2026-07-17
fix: missing nvidia dep in base

ConnorNeed:
Caught a missing dependancy for video streaming in our base image. Our dev image had it which is why it took so long for me to notice


ConnorNeed:
Dockerfile syntax is a pain :(

The power of a nice lint test. 

We won't merge until it passes the lint test.

codeflight1 (approved):

PR #173, opened 2026-07-18
feat: move video presets to input_node config

codeflight1:
Moves video presets to the `input_node` config for a single source of truth between webUI and joystick carousell controls. Draft until I test on-rover.


ConnorNeed:
Overall not a huge fan of this architecture. I believe the client should be allowed to set any configuration it wants without it necessarily being saved. IMO this node is already handling more than it's fair share of logic and this further grows it into a untracable mess of spaghetti code. Parameters should not be used as state tracking.

My thoughts would be to either: seperate the preset handler into it's own class or just maintain a list of presets that can be read by clients with the client then using them as it see's fit.

During testing I also found it very unreliable

codeflight1:
Agreed, I've updated to a new architecture with a separate node responsible for presets, which is much cleaner imo

src/Cameras/video_streaming/config/presets.yaml
ConnorNeed, line 1: Probably want a mast cam too

ConnorNeed (approved):
Looks good. I will see about testing tonight but no promises

src/Cameras/video_streaming/config/presets.yaml
codeflight1: I think so, will talk to Jack about needed presets tomorrow

PR #176, opened 2026-07-21
Remove discovery server

ConnorNeed:
Removes the discovery server

Fixes the eef direction being swapped (It was wired backwards compared to what it was.


ConnorNeed:
Merging in since Jack is coming in to practice tonight and I want the most realistic setup for him. Let me know if there's anything you want different and I can open a new PR

